PR #7, opened 2015-01-14
Modified mock factory interface with profile components

aphearin:
This is a rather trivial pull request. The purpose is primarily to serve as a sanity check on my github configuration after deleting the github.com/halotools account.


aphearin:
Everything looks kosher. Rejecting trivial pull request.

PR #19, opened 2015-02-04
Update scipy to a requirement

### LINKED ISSUE #18: Tests failing on travis due to missing scipy
The problem mentioned in #17 is now mostly fixed (a few of the builds still have it, but a fix is in the works for that from the astropy_helpers side).  But now that the tests are actually running on travis, a different problem has showed up.  To see an example, take a look at one of the failing builds from https://travis-ci.org/astropy/halotools/builds/49241697 .  There, you'll see a bunch of errors like ``ImportError: No module named scipy.special``. 

What's happening there is that the default builds do *not* include scipy.  This is because, in astropy, we have a general guideline that scipy is not a strict requirement, but rather it is necessary for certain things (like model fitting) that use scipy.  So any test that invokes functionality requiring scipy is skipped if scipy is not installed, but run if it is.  For an example of this, see https://github.com/astropy/astropy/blob/v1.0rc1/astropy/modeling/tests/test_fitters.py - you'll see a block at the top that defines the ``HAS_SCIPY`` variable, and then tests below that require scipy have decorators that look like ``@pytest.mark.skipif('not HAS_SCIPY')``.

As it stands, however, even that will not work for halotools - this is because there are at least some places where scipy is imported at the top level.  (E.g., https://github.com/astropy/halotools/blob/master/halotools/halo_occupation.py imports a variety of things from scipy).  In astropy, the convention (where possible) is to have scipy imports always *inside* a function/method, so that the error only happens when that functionality is used, rather than when the package is imported.  So then stuff that requires scipy fails, but everything else can still succeed.

So there are two different solutions:

1. Update halotools both to skip the scipy-requiring tests *and* move all scipy imports inside functions.
2. Just take the stance that scipy is a dependency of halotools.  ``setup.py`` should updated to add scipy to the ``requires`` list, and travis can be updated so that scipy is always installed (both of those are pretty much one-line changes).

I can set up option 2 myself if thats the way you want to go, @aphearin, but of course someone more familiar with halotools would have to do option 1.

eteq:
This updates halotools to have scipy as a requirement.

It *should* fix #18 , @aphearin, but we'll see what travis says...


aphearin:
@eteq - looks like the Travis build still failed. Should we merge or reconnoiter?

eteq:
``ImportError: No module named mpi4py`` is now what it's failing with, @aphearin (the best jobs to check are the ones that have ``	SETUP_CMD='test'``).  Should mpi4py also be a dependency, or is that more of an optional thing?

aphearin:
Ahh, now *that* is a genuine halotoos bug. mpi4py should definitely be an optional dependency. This must be in Duncan's domain. @duncandc - can you fix this?

Here's what I recommend. Instead of importing mpi4py, instead do this:
try:
    import mpi4py
    has_mpi4py_global_variable = True
except ImportError:
    has_mpi4py_global_variable = False

Then, right before for your routines that use MPI, insert a simple control flow:
if  has_mpi4py_global_variable==False:
    raise ImportError("Tried to use MPI without having mpi4py installed")

Make sense? I think this will resolve the problem (until the next possible next one, that is).

duncandc:
This is in fact in my domain.  This should be an easy fix.

PR #23, opened 2015-02-06
comment-out the travis builds checking numpy versions other than 1.9

### LINKED ISSUE #21: Travis build still failing for numpy < 1.9 builds
As @aphearin  noted in #20, some of the travis builds are still failing with an error about ``TypeError: run_tests() got an unexpected keyword argument 'repeat'``

In fact, this is what d62c292ec3a067e25e462e19bd60463284d7a13d fixed by switching most of the builds to use numpy 1.9 . The underlying problem was that the astropy version installed by conda is only available for numpy < 1.9 as astropy v0.4.1 (which is apparently incompatible due to this bug you noticed).  But those last three builds are there specifically to check versions of numpy < 1.9 ... which of course now fail.  

At some point the plan is to fix ``astropy_helpers`` to make this work again... But until then these builds will fail.  Perhaps the best thing is to just comment those out for now and put them back in when/if astropy_helpers is updated?

eteq:
This closes #21 by just not running the numpy builds for versions < 1.9. Hopefully the travis tests will start passing once this is in.

This is probably best considered a work-around, because there are some subtle bugs that can creep in if you don't test older versions of numpy.  Unless of course you don't mind just requiring numpy 1.9 (or at least saying "untested/not guaranteed for older versions) - then there's no need to test them at all.

cc @aphearin

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,12 +47,13 @@ matrix:
           env: SETUP_CMD='test'
 
         # Try older numpy versions
-        - python: 2.7
-          env: NUMPY_VERSION=1.8 SETUP_CMD='test'
-        - python: 2.7
-          env: NUMPY_VERSION=1.7 SETUP_CMD='test'
-        - python: 2.7
-          env: NUMPY_VERSION=1.6 SETUP_CMD='test'
+        # currently these are commented out because the conda copy of astropy is only up-to-date for numpy 1.9
+        #- python: 2.7
+        #  env: NUMPY_VERSION=1.8 SETUP_CMD='test'
+        #- python: 2.7
+        #  env: NUMPY_VERSION=1.7 SETUP_CMD='test'
+        #- python: 2.7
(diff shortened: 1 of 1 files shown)


aphearin:
Well, I don't know. 1.9 is the most recent version, so it's a little uncomfortable to make this a hard requirement. But then again, literally any modern python distribution package (e.g., conda) will keep users up to date with numpy, so maybe this is not such a big deal to require. My vote is to just require v1.9 unless you consider this very misguided, @eteq .

eteq:
That sounds fine for now.  Note that it still *might* compatible with older numpys,  just you won't be testing to make sure.  My experience has generally been that if it works with 1.9 it will with 1.8 and 1.7.  1.6 is harder because a number of features were added in 1.7 that you might actually use (1.8 and 1.9's new features have thus far not been "must-have" for anything I've done, although I think there were some important bug fixes).

The failures now in this PR seem to be unrelated, so I'm going to go ahead and merge.

PR #24, opened 2015-02-17
Test Pull

kevinhuangtan:
First pull attempt. Testing setup for workflow.


aphearin:
This is a good start. The workflow appears to be correct: you pushed to your personal fork and then submitted a PR from there. 

The code needs to be restructured though. Let's see how it looks after making the following changes:

1. Your wrapper should be located in halotools/halotools/mock_observables/pair_counters/ and the files should be given a more informative name. We can't have files just called "wrapper.pyx". 

2. Manodeep's C code should be stored in a directory called halotools/cextern/sinha_pair_counter/

3. There are a large number of files that should be excluded. Briefly, the only files you should include are the absolute bare minimum of files needed to count the pairs. Below is a (possibly incomplete) list of files that should deleted before your next PR.  Be sure to check that the code still compiles, wraps, and operates properly after removing them. 

* halotools/mock_observables/sinha_pairs/bins
*halotools/mock_observables/sinha_pairs/cmassmock_AbM_swot_scatter0.10dex_All_V_PEAK_Zspace.ff
* halotools/mock_observables/sinha_pairs/convert_ascii_to_fastfood.pro
* halotools/mock_observables/sinha_pairs/io.h
* halotools/mock_observables/sinha_pairs/notebooks/*
* halotools/mock_observables/sinha_pairs/openmp_speedup_Mr19.png
* (There are other .png files, all of which can be removed)
* halotools/mock_observables/sinha_pairs/output
* (All .pro files can be removed)
* (All .ff files can be removed)
* halotools/mock_observables/sinha_pairs/test_pair_counter.py
* (All .pdf and .txt and .eps files can be removed)

PR #25, opened 2015-02-17
Deleted extra files and place files in correct directories

kevinhuangtan:
I deleted all the files from sinha_pair_counter that I could  delete from the directory without the Makefile failing. If I delete io.h or any other file, I will have to change the source code in DD.c to accommodate this. Let me know if you would like me to do so, or to leave the code as is for now.


duncandc:
This looks good.  I will merge this and give it a whirl.

PR #28, opened 2015-03-02
Removed ktan wrapper from repo

aphearin:
The sinha wrapper was not installed properly when merged, which was leading to a fatal error when attempting "python setup.py install". I removed the wrapping code entirely as a temporary fix, but it will need to be included back in the repo when it works. @duncandc


aphearin:
Note that there are now two test failures in test_count_pairs @duncandc

PR #34, opened 2015-03-23
Latest version

kevinhuangtan:
If libcountpairs.so is in the directory of sinha_pair_wrapper the function cy_countpairs will work. Right now the build it is unable to link to cextern


aphearin:
@kevinhuangtan - can you flesh out the operating instructions a little more. In particular - how is libcountpairs.so created? Exactly what compiler command, and what flags need to be thrown, to compile Manodeep's code in the correct way your code is set up for? Is it *just* libcountpairs.so that needs to be in the same directory as sinha_pair_wrapper, or does Manodeep's entire library need to be located in the same directory as your wrapper?

aphearin:
* Your cextern directory only has pre-compiled binaries. Instead, what is necessary is the (possibly modified) version of Manodeep's Makefile, and all .c, and .h files needed to compile Manodeep's code (and only those files)

* It looks like you did not include the necessary .pxd file in this PR. 

* You seem to have hard-coded a dirname that is particular to your own machine.

PR #44, opened 2015-04-16
update to helpers v1.0.2

eteq:
While looking into duncandc/halotools#1, I noticed an odd error message that was being thrown while trying to do ``python setup.py build`` about the astropy_helpers not being valid.  This PR fixes that by updating astropy_helpers to point to the latest version.  @aphearin or @duncandc , the way I did this was:
```
<starting from inside halotools root>
cd astropy_helpers
git fetch
git tag 
<this lists the tags, which showed that 1.0.2 is the latest>
git checkout v1.0.2 <"git checkout master" is usually fine too, as it's generally stable>
cd ..
git add astropy_helpers
git commit -m "updated to helpers v1.0.2"
```
it's often small enough changes that you can just do it directly in master instead of bothering with a PR.  Probably not critical to do unless something you know is important is in there, too (except right before a release - in that case you always want the most up-to-date).  I did it in a PR here just so I could mention it here (and to fix whatever bug it was that was causing the error message)


eteq:
@aphearin - the test failure is "normal", right?  That is, it's not created by this PR?

aphearin:
Yes, this test failure is because of other pair counters written by @duncandc that will soon be removed from the repo.

PR #52, opened 2015-05-10
Merged in updated astropy-helpers 

aphearin:
So astropy_helpers had gotten out of sync somehow. I'm pretty sure this was my own doing, though even in hindsight I'm not sure how to have recognized how it was happening. It is a big of a nuisance that we have an entirely separate git history located in a sub-package, because it means we need to remember to manually check the astropy_helpers git history before committing to master. 

**Bottom line**: with these changes, all of the changes made by @duncandc seem to have paid off: from within my bare bones virtualenv, the code passes all tests and raises no errors! Woohoo!


aphearin:
Well, durn, Travis is more finicky than my virtualenv.

PR #55, opened 2015-05-12
Changed astropy helpers to v1.0.2

aphearin:
As per the suggestion by @eteq - I used git checkout v1.0.2 from within the astropy helpers sub-package. **This placed the astropy helpers package located within Halotools into a 'detached HEAD' state.** Added this detached head state to the master branch of Halotools.


eteq:
@aphearin - that's normal for this situation.  It's a detached head because the tag points to a commit that's not the *head* of an existing branch. So if you were to go and make more commits from there, they'd be lost if you don't create a new branch. But for this situation, where you're going back to commits already existing in the commit tree without actually making changes, it's not a problem.

eteq:
but it is a scary warning, I agree... sometimes git can be alarmist...

Issue #58, opened 2015-05-13
cython building error


duncandc:
@aphearin @eteq 

I get the following error when building a clean clone of halotools once I add a cython module:

"Cython must be installed to build from a git checkout."  

I do in fact have cython installed.

aphearin:
Weird. Could you try adding 'cython' to the install_requires field of setup.py?

duncandc:
no dice.  same error.

eteq:
@duncandc - that is strange.  Is it possible you've got conflicting python installations?  Like it thinks cython isn't installed because it installed it in a place it can't see?

Can you push the commit in question up to some branch somewhere so that I can separately see if I get the same error?

duncandc:
@eteq and @aphearin 

see merge latest request.

aphearin:
Awesome. I'll test it asap.

aphearin:
Ok, so I checked out the mock_obs_devel branch that @duncandc has up on his fork. The following works:

python setup.py build
from halotools import mock_observables

However, this seems to be because the __init__.py files are empty. I tested this with the following changes:

To halotools/mock_observables/__init__.py, I added:
    from .pair_counters import *

And to halotools/mock_observables/pair_counters/__init__.py, I added:
    from .sinha_pairs import *

With only these changes to the code, when I try to import the code I get:
    ImportError: No module named sinha_pairs_wrapper

This is a little perplexing, because the code apparently does compile and build. Any ideas @eteq ?

duncandc:
I have noticed this odd import behavior.  The following works:

from halotools import mock_observables.pair_counters.sinha_pairs as pairs

Then you can count pairs with 
pairs.npairs()

aphearin:
Huh. That's puzzling. Let's see if @eteq has experience with something like this before we merge.

aphearin:
Also, I have not handled multiple simultaneous PRs before. Since @duncandc  submitted this PR, I have updated the master branch at least once, I think twice, with significant changes to the documentation. @eteq - is it safe to just use the github tools to merge directly, or should I locally clone from the branch on @duncandc 's fork, merge locally, and submit a new PR? What do y'all do in Astropy?

duncandc:
I can do it on my machine.

aphearin:
I don't mind doing it at all. I just want to know whether it's necessary to go through that rigamarole.

eteq:
[...]

PR #59, opened 2015-05-14
Add Simulation image to top bar

eteq:
This alters the halotools docs to add an nbody simulation image to the top bar, as @aphearin and I discussed offline.  You can see how it looks when generated at http://eteq.github.io/halotools/ .

I can't claim credit for figuring out how to make this work - I copied the trick of a customized css that inherits from the astropy style from gammapy (thanks @cdeil!)

@aphearin - I actually cut out a strip from the file you pointed me to, rather than the whole thing - the whole thing was 1.5M, and that would make all visitors to the docs page have to download that whole file.  The strip I cut out here is a more manageable 150K.  If there's some other chunk of that image you'd like to chop out, feel free to PR against this branch or otherwise send it to me.  Also, this image should probably be credited somewhere.  Where do you want me to put that? (or, again, feel free to issue a PR against my add-haloimg-topbar branch).


eteq:
@aphearin - as per our e-mail discussion, I've added mention of Ralf and KIPAC, right now on the front page because I didn't see a credits page in the documentation.  I can easily put it elsewhere if you want, though. (Or this can be merged and that changed later).

aphearin:
I think that's fine for now. Ralf didn't seem too worked up about this. Feel free to either go ahead and merge, or charge me with doing it, subject to the following caveat. I could handle the merge myself, but before doing so I would want to know the answer to my question in #58 about how GitHub handles multiple PRs, some with a time lag.

eteq:
@aphearin - re: your question in #58 and how it pertains to this PR: if you merge this now it looks like it won't conflict at all with #58 because none of the same files are altered by these PRs.  So this should be safe to merge right now without affecting anything else, but I'll leave it to you to do that to make sure you're ok with it.

aphearin:
Rebase. Got it. I read about this command in a git tutorial months ago and have never used it, but this is exactly the functionality I was asking about. Thanks for the tip. I'll read up on it.

PR #61, opened 2015-05-14
Updated install requirements

aphearin:
Code successfully installs both on my global system as well as in a virtualenv. I think we're in business for pip --pre-release registration.


aphearin:
Code successfully installs both on my global system as well as in a virtualenv. Next step is to upload the code to Pypi for --pre-release installation.

PR #63, opened 2015-05-14
added sinha pair counter back into mock obs

duncandc:
@eteq @aphearin  This now compiles and imports on my system with the most recent version of master.

Someone else give it a try and let me know.


halotools/mock_observables/pair_counters/sinha_pairs/sinha_pairs_wrapper.pyx
```diff
@@ -0,0 +1,192 @@
+from countpairs_pbc cimport countpairs_pbc as countpairs_pbc
+from countpairs_nopbc cimport countpairs_nopbc as countpairs_nopbc
```
eteq: is there a reason for the ``as`` parts?  it looks like this is exactly the same if you left off the `` as ...`` parts.

eteq:
@aphearin - I tried the thing you mentioned in #58 of adding the ``import *`` statements to the new ``__init__.py``'s, and I did *not* get the error you did about ``ImportError: No module named sinha_pairs_wrapper``.  Is it possible you weren't importing the built version or something?  I.e., if you did the import directly in the base halotools directory instead of the ``build/lib...`` directory. 

@duncandc @aphearin - I also tried this just now on my ubuntu machine at home (which uses gcc instead of clang), and it doesn't compile, for the same reason it didn't in the old version: something with the ``restrict`` keyword in ``countpairs.h`` makes gcc unhappy. Will try to figure this our later, but if either of you know what ``restrict`` means here that might be helpful...

aphearin:
A very recent development is that, after lots of sweat developing our own pair counter, I think we may not be using Manodeep's pair counter after all. It's quite a headache to get it to robustly compile, and nearly impossible to edit for alternative purposes. Let's consider this a closed issue for now, coming back to it only if the native numpy + cython pair counter duncan and I recently wrote does not turn out to be workable.

PR #75, opened 2015-06-11
Added in random downsampler

aphearin:
See halotools.utils.array_utils for the randomly_downsample_data method


aphearin:
Ask and you shall receive, @andrew-zentner

PR #81, opened 2015-06-21
Added Generalized CAM model

aphearin:


aphearin:
@duncandc - I use a slightly different algorithm for implementing CAM scatter than you. The new method is still based on Pearson rank-order correlation coefficients, but it is entirely robust to the dynamic range and PDF of the secondary galaxy property. I have also implemented variable scatter as a function of stellar mass, or whatever the primary galaxy property is. 

The ConditionalAbunMatch docstring is comprehensive and include a fairly extensive set of examples. There is also a pre-loaded model called Campbell15. The testing I have done is relatively broad, so at this point the only way to effectively find bugs is to just start using the class. 

The main problem I'm having is runtime: it takes me about 20 seconds to populate Bolshoi with Campbell15. This is still fast enough to be quite useful for exploratory science, because of how much flexibility there is in the class. But obviously this is unacceptable for likelihood analyses. I have not profiled the code yet, so I don't know what's taking so long, other than that ConditionalAbunMatch is responsible. It would be useful to hear what runtimes you've been able to get with your own code. In general, notes on API, clarity, and performance are most welcome.

aphearin:
@duncandc - here are the relevant links to the docs:

http://halotools.readthedocs.org/en/master/api/halotools.empirical_models.ConditionalAbunMatch.html#halotools.empirical_models.ConditionalAbunMatch

http://halotools.readthedocs.org/en/master/api/halotools.empirical_models.Campbell15.html#halotools.empirical_models.Campbell15

PR #92, opened 2015-07-02
Added new module dedicated to halo/galaxy catalog manipulation 

aphearin:
See the SampleSelector class in utils/table_utils.py


aphearin:
@andrew-zentner and @duncandc - you will both likely find the following method quite useful more or less immediately: utils/table_utils.SampleSelector.split_sample. This returns sub-samples of an Astropy table according to percentile-based subdivisions. Provided that the input table is of sufficient length, the method supports arbitrary subdivisions. I've taken some care to properly deal with edge cases, so this *should* return robust results. 

You can find the docs here:

http://halotools.readthedocs.org/en/master/api/halotools.utils.SampleSelector.html#halotools.utils.SampleSelector

PR #93, opened 2015-07-03
major update to mock observables

duncandc:
added FoF groups
added fancy pair counting, (objective weighting functions, and s and mu binning)
mock redshift survey creation (distant observer and ra,dec,z)
more TPCF functions
re-reorganized some of the cython pair counting modules


aphearin:
This looks like a huge leap forward in functionality, @duncandc . I'm excited to start playing with this, but I'm having some new compile problems that I have not had before. @eteq - could you let me know if you've seen the following error before?

So the test suite passes with flying colors, which includes a bunch of tests of the new code. However, I cannot import mock_observables. What *used* to work was executing the following two lines from the halotools root:

python setup.py build
python setup.py build_ext --inplace

Then I had no problem importing mock_observables. Now when I try to compile the code with the above lines, the first line causes no problems, but the second line throws the following error:

error: [Errno 2] Could not find C file halotools/mock_observables/pair_counters/cpairs/cpairs.c for Cython file halotools/mock_observables/pair_counters/cpairs/cpairs.pyx when building extension halotools.mock_observables.pair_counters.cpairs.cpairs. Cython must be installed to build from a git checkout.: 'halotools/mock_observables/pair_counters/cpairs/cpairs.c'

This a new error that has not happened before, at least not on my machine. Now when I try to import mock_observables, I get an ImportError: no module cpairs. 

It's also worth noting that if I cd into build/lib.macosx-10.5-x86_64-2.7, then I can import mock_observables just fine. 

It's

duncandc:
run "git clean -dfx", then run "setup.py build_ext --inplace".

You can now not import mock observables from within the halotools package.  It must be imported from the halotools root directory.  I have already discussed this "feature" with @eteq .

aphearin:
Sorry, I still don't understand. My confusion is over the distinction you
make between "within the halotools package" and "from the halotools root
directory". Can you be more explicit? I just don't follow.

duncandc:
you have to import as e.g. from halotools.mock_observables.clustering import tpcf

you can't import e.g from clustering import tpcf from within the mock_observables directory.

aphearin:
Ahh, ok, got it. That does indeed work. So what tripped me up is that the following:

$ git clean -dfx
$ python setup.py build
$ python setup.py build_ext --inplace

That fails. Good to know.

PR #99, opened 2015-07-14
work around for igraph package naming issue.

duncandc:
added a kind of ugly work-around to deal with the issue that there are two packages called igraph.  I am not sure how we want to deal with this in the long run.

@aphearin do you want to see if this solves your problem?

### DIFF
--- a/halotools/mock_observables/groups.py
+++ b/halotools/mock_observables/groups.py
@@ -16,6 +16,10 @@ try: import igraph
 except ImportError:
     igraph_available=False
     print("igraph package not installed.  Some functions will not be available.")
+if igraph_available==True: #there is another package called igraph--need to distinguish.
+    if not hasattr(igraph,'Graph'):
+        igraph_available==False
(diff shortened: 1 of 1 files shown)


aphearin:
The only problem I see with this is that it seems pretty plausible that the alternate version of igraph will end up having a sub-package called Graph. Upon inspection of both versions of the igraph package, the __license__ for the two is quite different. For the one we don't want, __license__ is just the three-character string "MIT". For the one we do want, the __license__ is quite long, and includes the name of the package author, who is Hungarian. I think using "'Nepusz' in igraph.__license__" as a trigger would be a more effective way to protect against this.

aphearin:
So I don't know how to turn off Markdown features in GitHub. So the boldface type in that last post should actually be a double-underscore, both leading and trailing.

aphearin:
Don't worry about fiddling with this anymore. I'm going to merge this and make the change myself later.

PR #100, opened 2015-07-15
Param dict decoration

aphearin:
Eliminated need for component models to use update_param_dict and input_param_dict mechanism by the model factory's use of function decorators


aphearin:
@andrew-zentner this simplification is now being merged into master

PR #107, opened 2015-07-24
Thorough overhaul of sim_manager sub-package

aphearin:
Major revision to all aspects of the sim_manager. 

1. The behavior of ascii reading, catalog management, and halo catalog objects are now entirely decoupled. 
2. There are now 20 pre-processed halo catalogs (5 different simulations with 4 redshifts each) stored as hdf5 files on my website, any/all of which can be downloaded and stored to the Halotools cache using sim_manager. 
3. There is also support for use of an external drive as a secondary cache, as the full collection of halo catalogs occupies ~5 Gb.
4. Most halo catalogs (currently 14 of 20) are accompanied by a randomly selected downsampling of ~1e6 dark matter particles. 

One of the main motivators for the overhaul was to simplify the distinction between a halo catalog and mock galaxy catalog. All columns of a halo catalog have key names that begin with 'halo_', greatly simplifying the bookkeeping of making mocks with a composite model, where some components get passed a halo catalog, others the partially-built mock galaxy catalog. This change required very substantial changes to be made to the function signatures of most composite models. 

Through the ModelFactory's use of the _update_param_dict_decorator, parameter passing from the composite down to the components now happens automatically, eliminating the need for component models to use the input_param_dict mechanism. 

There are now also tests in the suite that marked with APH_MACHINE. These tests are intended to be specific to my own personal cache configuration, but this should be confirmed on an independent machine before merging.


aphearin:
@duncandc or @eteq - could one of you do me a simple favor? Checkout some new branch starting from astropy/halotools/master, bring in the changes from this branch, and then execute the test suite and let me know if there are failures. I have attempted to set up a test marking system such that there are tests which should only run when the suite is executed from my own personal machine, but I would like to confirm with an independent machine that I have done this correctly before I merge.

PR #123, opened 2015-08-12
changes and updates to clustering module.

duncandc:
This has some bug fixes for the clustering module, and updates to the doc strings.

This merge had some conflicts with the master, arising from doc strings edits by @aphearin , which I resolved, but this may make the merge into the halotools master complicated.

### DIFF
--- a/halotools/mock_observables/clustering.py
+++ b/halotools/mock_observables/clustering.py
@@ -17,6 +17,7 @@ __author__ = ['Duncan Campbell']
 
 np.seterr(divide='ignore', invalid='ignore') #ignore divide by zero in e.g. DD/RR
 
+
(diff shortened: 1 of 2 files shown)


aphearin:
This looks fine to me. No merge problems, and the docs build without complaint.

aphearin:
Well, I tested that the docs built locally on my machine before I merged. However, ReadTheDocs fails to build them. What a pain. RTD can't even import the project as of this latest PR. No idea why. Will look into it.

aphearin:
Problem solved itself. Sometimes RTD just like to think about it for a while, issue a succession of build-failures, and then agree to comply after sufficient contemplation.

Issue #77, opened 2015-06-12
Add support in mock factories for models returning more than one property


aphearin:
For example, a component model may simultaneously return a disc scale length and a bulge-to-disc mass ratio.

Issue #149, opened 2015-09-01
extraneous halo boundary


aphearin:
Once the profile modeling has been overhauled, delete the extraneous halo_boundary definition in the model_defaults module

aphearin:
Done.

Issue #144, opened 2015-08-20
Enforce consistency between halo profile models and halo prof param models


aphearin:
Currently, nothing at all is done to ensure that a given choice of halo boundary definition is consistent with the conc-mass relation used by the model.

This issue was imported from the aphearin repo and may be obsolete.

Issue #143, opened 2015-08-20
Enforce profile model consistency in composite model


aphearin:
If a composite model is built from two (or more) component models that happen to have profile parameters with one (or more) common keys, then it needs to be enforced that these parameters correspond to the same profile class. Otherwise, buggy behavior will result in the composite model and mock factory, because the way the composite model lookup tables are built is that the first matching key that is found, the table is built from that component model.

This issue was imported from the aphearin repo and may be out of date.

PR #158, opened 2015-09-11
Overhaul of the 1-halo term 

aphearin:
This PR introduces isotropic Jeans velocities as an HOD model feature. 

The class and module structure of the profile modeling has been re-written from the ground-up, now based on multiple inheritance (orthogonal mix-ins only). 

All component models now declare what galaxy_table columns they will assign, and overwrite the table columns, as suggested by @andrew-zentner . This restructuring required a relatively major overhaul of both the model and mock factories, which are now far more general and support a user-defined calling sequence, enabling, for example, HOD-style models with distinct red/blue SMHMs as in Tinker et al. 2013.


aphearin:
PR resolves #77, #8, #144, #143, #149.

PR #174, opened 2015-10-02
Convenience function for generating points in NFW phase space

aphearin:
mc_generate_phase_space_points function added to NFWPhaseSpace class. 

This is purely a convenience function for returning a data table of points in NFW phase space.


aphearin:
@duncandc - I'm not sure, but this might be of use to you for estimating the impact of Poisson noise on estimates of the velocity dispersion of the particles in the halo core.

PR #181, opened 2015-10-14
Rewrite of the pair-counting engine

aphearin:
As of this PR submission, the current branch is **NOT READY TO BE MERGED**. This is a code-review request. 

The underlying engine that drives the two-point function calculations has been rewritten. This is now being handled by the FlatRectanguloidDoubleTree class. The most important change is that the size of the cells into which the simulation subdivided is now *mostly* decoupled from Rmax and Lbox, and all three dimensions are now being treated entirely independently. The currently pushed version has only been implemented for the 3d, unweighted pair-counter. I anticipate the generalizations to the MCF, jackknife, 2+1 dimensional counters to be *mostly* straightforward, but will involve careful attention to details. 

In a little more detail, the API is such that an *approximate* cell size is passed to FlatRectanguloidDoubleTree. The constructor uses this approximate guess and chooses the closest cell size that meets the following constraints:
    1. The cell size in each dimension must evenly divide the size of the simulation box in that dimension
    2. The cell size must at least exceed Rmax in each dimension
    3. There must be at least 3 cell-lengths per dimension

I find that I can improve typical pair-counting for SDSS-like samples in Bolshoi by a factor of 2-3, and in Multidark by 2-100x, depending on the regime. However, I have not yet found a universal organizing principle for how to choose the optimal cell size, and the performance depends fairly sensitively on the choice. My current solution is to use Lbox/10 as the default, which I find always gives reasonable results, and to be very explicit in the docstring that some experimentation should be done for performance-critical applications. 

CC @duncandc , @surhudm

### DIFF
--- a/halotools/mock_observables/double_tree_clustering.py
+++ b/halotools/mock_observables/double_tree_clustering.py
@@ -1,0 +1,327 @@
+# -*- coding: utf-8 -*-
+
+"""
+functions to calculate clustering statistics, e.g. two point correlation functions.
+"""
+
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+####import modules########################################################################
+import sys
+import numpy as np
+from math import pi, gamma
+
+from .pair_counters.rect_cuboid_pairs import npairs
+##########################################################################################
+
+
+__all__=['tpcf']
+__author__ = ['Duncan Campbell']
+
+
+np.seterr(divide='ignore', invalid='ignore') #ignore divide by zero in e.g. DD/RR
+
+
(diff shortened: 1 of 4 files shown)


aphearin:
Two more comments about the changes implemented to the engine:
    4. The length of the cells of data2 are required to evenly divide the length of cells of data1 in each dimension
    5. I **pre-shift** each sub-volume of points before passing the pair of sub-volumes to cython. In this way, I always compute pure Euclidean distances within the tight double-for loop, and never call the ``_pbc`` functions.

PR #253, opened 2015-11-24
update travis builds for conda and travis changes

eteq:
This updates the ``.travis.yml`` file to hopefully get travis working again - ``anaconda`` recently made some changes that were causing the tests to not run correctly.

Moving forward, we'll probably want to do what astropy/package-template#140 is suggesting, but I'm not sure how to do that you, so probably best to get this in now, and then when astropy/package-template#140 is merged, pull that out and into halotools.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,22 @@ python:
     - 3.3
     - 3.4
     # This is just for "egg_info".  All other builds are explicitly given in the matrix
+
+# Setting sudo to false opts in to Travis-CI container-based builds.
+sudo: false
+
+# The apt packages below are needed for sphinx builds. A full list of packages 
+# that can be included can be found here:
+#
+# https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-precise
+
+addons:
+    apt:
+        packages:
+            - graphviz
+            - texlive-latex-extra
+            - dvipng
+
 env:
     global:
         # The following versions are the 'default' for tests, unless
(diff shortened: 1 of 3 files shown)


eteq:
@aphearin - OK, I think this is now actually working.  There are still some failures, but I believe they are *real*.

@duncandc @aphearin  - note that this required a small modification to ``setup_package.py`` to use the numpy headers with a special incantation of just the string ``'numpy'`` instead of the ``numpy.get_include()`` function.  You had no way of knowing that this is a requirement to get the ``egg-info`` command to work, but for more on this, see astropy/astropy#4357 .  Or just ignore it and accept that this PR fixes it ;)

As for the "real" failures, it's basically two issues:
* A few doc build failures which look to be some typos and duplicated API pages. 
* All of the py 3.x builds are failing.  I could change this to "allowed failures" if you don't want to tackle this in the near term, @aphearin.  But note that 3.x compatibility is a fairly important piece that typically only gets harder the longer it gets put off...

I'd suggest this get merged as-is to get at least some of the tests going, although if py 3.x should be marked as allowed fail, I could put up one more commit to do that.

aphearin:
Great work, @eteq. I'd like to leave 3.x optional for now and tackle this in one large sprint after sending the code out for beta testing, but before the first official release. The docs duplication I will address this week.

aphearin:
Please feel free to add that one last commit and merge.

eteq:
Ok, updated to allow py 3.x failures.  Assuming this works I'll go ahead and merge.

One other thing I forgot to mention above: this also adds the machinery to support reports from ``coveralls.io``.  If you go to https://coveralls.io/github/astropy/halotools you'll see a little report that tells how much of the code is tested, and where.

Presumably this isn't terribly fair because some of it is tested for scientific accuracy in slower ways that you don't want on travis... But it's at least a decent guide.  We can also adjust the settings so that it actually tells you if a new PR increases or decreases coverage.  (Feel free to discuss this in Slack if you're interested, @aphearin).

eteq:
Looks good - now once the sphinx stuff is fixed, the build should come in passing.  Will merge now, and add an issue about how to bring back 3.x once it's working.

PR #272, opened 2015-12-01
Added a few blank lines where sphinx was complaining 

aphearin:
These were in the marked_tpcf docstring. Now checking to see if this resolves the Travis failure.


aphearin:
Note to @duncandc and @eteq - Travis tests now pass. Let's try to keep it this way from now on.

PR #245, opened 2015-11-24
Make autosummary tables work

eteq:
As recently discussed at length with @aphearin on slack, it turns out there's a bug in `sphinx`'s autosummary that was causing many of the associated problems with the docs.  This PR addresses that by adding a small sphinx extension that just fixes the bug 

The extension looks complex, but it's really just copied-and-pasted from sphinx, with the one change being on line 28: ``if not documenter.check_module()`` -> ``if documenter.options.members and not documenter.check_module()``. 

This also fixes up `phase_space_models_functions.rst` to be compatible with the extension - other docs that are supposed to have `autosummary` tables will need to be switched to follow this form (or similar).  

This also makes a few changes in the docs, most visible of which is the addition of a section for the Reference/API.  @aphearin, I can back that out if you don't want that in, but it might make sense given that this autosummary stuff is now linking directly to that.

PR #275, opened 2015-12-01
Replace 245 with updated helpers

eteq:
Ironically, a "real" fix to #245 was implemented independently (and in a cleaner way) by astropy/astropy-helpers#172 at abount the same time as I did #245 .  So this PR removes the workaround from #245 and updates the helpers to point to the latest master (which *does* work).

It might make sense to wait on merging this until there's a new helpers release - one should come very soon along side Astropy v1.1, and it would make sense to use a release here instead of pointing to master (not necessary, but might be helpful).


aphearin:
@eteq - there now appears to be a `release-1.1` branch on astropy-helpers. Does this mean that the official v1.1 is released? If so, should we just close this PR and update the helpers to 1.1?

aphearin:
See #294

PR #276, opened 2015-12-01
switch catalog_manager exceptions to raising an exception instead of strings

eteq:
I'm pretty sure the use if ``raise`` with a string wasn't intended and is a holdover from before there was a `HalotoolsError `.  But at any rate, raising strings doesn't work anymore anyway, so this fixes that.


eteq:
Oops, I now see the diff is full of whitespace things that my editor does automatically.  But I swear the only substansive changes are like those at https://github.com/astropy/halotools/pull/276/files#diff-2c1e5123431e87ec2371aa5052b15b96L18

PR #297, opened 2016-01-11
Ground-up rewrite of the sim_manager sub-package

aphearin:
Almost all of the previous code in this sub-package has been deleted wholesale. This PR introduces many thousands of new lines of code. Test coverage for the code in this sub-package is 90%. 

Summary of changes:

1. The full functionality of Halotools is now available for use with halos in any N-body simulation identified with any halo-finder. You need only pass a sequence of numpy arrays to the UserSuppliedHaloCatalog class, providing a minimal set of metadata (Lbox and particle_mass), and the resulting instance can be used to populate mock catalogs with the usual Halotools model factories. UserSuppliedHaloCatalog instances can be stored in the HaloTableCache, or not, depending on user preference, as described below. If users choose to cache their personal halo catalogs, from then on they can reload their catalog with the usual one-liner: >>> halocat_obj = CachedHaloCatalog(simulation_nickname, halo_finder_nickname, redshift, version_name).

2. There is an all-new, object-oriented HaloTableCache that the user can choose to use to keep track of halo catalogs. The HaloTableCache provides wrapper-behavior around a simple, human-readable ASCII file, ‘halo_table_cache_log.txt’, stored in $HOME/.astropy/cache/halotools/. This is also the default location where cached halo catalogs are stored, but the HaloTableCache permits halo catalogs to be stored in any location to which the user has read/write access. The user has the option to manually add and remove lines from this file, and/or the option to add and remove log entries using the HaloTableCache class, according to personal preference and circumstance.

3. Entries of the HaloTableCache are instances of the HaloTableCacheLogEntry class, which over-rides the python built-in comparison methods __eq__, __lt__, etc., which simplifies the search process for matching/duplicate entries. Log entries also have the ability to self-verify whether or not they are suitable to store in the cache, based on a large number of consistency checks between data and metadata.

4. Use of the Halotools caching system is now optional in every respect. When making mocks, computing observables, and analyzing catalogs, users are free to completely ignore the HaloTableCache, to use it exclusively, or anything in between. If the user does not wish to use the HaloTableCache at all, they need only write a single python function that returns catalog metadata and arrays of halo properties from their proprietary simulation. By passing the output of such a function to the UserSuppliedHaloCatalog constructor, the user has unrestricted access to the remaining functionality of Halotools.

5. There is now a lazy and memory-efficient TabularAsciiReader for processing ASCII of a dtype that is known in advance. This class is fully stand-alone, and is probably useful as-is outside of Halotools for “Big Data” applications where alternative options such as Pandas, Numpy or Astropy conk out. The TabularAsciiReader accomplishes this by making the following two hard assumptions about the structure of the ascii data: 1. lines of data appear consecutively following a user-defined header, 2. after the header, the first appearance of an empty line demarcates the end of the data stream, 3. the dtype of the desired columns is unchanging and known in advance. Although restrictive, the above assumptions are quite a common situation in present-day simulation analysis. The reader only needs to know the columns requested by the user, and only the dtype of the *desired* columns need be specified. Highly customizable row-cuts are applied on-the-fly as the file is read. Performance is tunable with an input chunk_size, set to 500 Mb by default so that any modern laptop can process a chunk at a time. Reduction-times with the default settings are roughly ~1.5 Gb/minute on my laptop. In the end, only the *final, reduced* catalog need fit in RAM; subject to that condition, the TabularAsciiReader can analyze an infinitely large file.

6. All of the above functionality is thoroughly documented on halotools.readthedocs.org, which has been considerably reorganized to improve discoverability of these features

The sim_manager sub_package is quite close to being ready for beta-release. The only remaining feature is to finish writing the class dealing with user-supplied collections of dark matter particles.

@andrew-zentner - Once this PR goes through, you and your students should reinstall the package. 

@eteq - This PR contains the ASCII reader we discussed, which expands the use-cases supported by Astropy readers. The only package needed to use it outside the standard library is numpy. 

@yyao - probably nobody has reduced more Rockstar catalogs than you. It would be useful to compare performance notes with your code and the TabularAsciiReader, and/or just get comments. Ok by me if you want to wait for the beta release to do that, but I don't expect to touch the TabularAsciiReader class again before beta-release.


aphearin:
@eteq - the Travis failures are all due to h5py import Errors. To remind you, I can't set h5py to be an installation requirement if I want the docs to build on RTD. So what I do now is only import h5py inside the appropriate the namespace. That solves the RTD problem. But this PR also uses h5py inside the test suite, so now Travis complains that my test suite raises unexpected exceptions. 

If I have that interpretation correct, this is entirely harmless for users and is purely a CI-inconvenience. First of all, please confirm. Maybe the workaround is to switch h5py import errors to be allowed failures (somehow?). Any alternative suggestions?

eteq:
(Leaving a comment here for posterity even though @aphearin and I discussed it already on Slack): I think it's *good* to test the no-h5py case and make sure it passes just so that we can be sure RTD is working, but most of the tests don't need to - so #298 has addressed that.

In the code itself, the thing to do for now is to remove the place where `h5py` is imported at the module-level (actually inside a class, but it still gets run on import). Then do something like this for places where `h5py` is needed in a test:
```
try:
    import h5py
    HAS_H5PY = True
except ImportError:
    HAS_H5PY = False

@pytest.skipif('not HAS_H5PY')
def test_which_uses_h5py(…):
    …
```

Issue #154, opened 2015-09-09
ReadTheDocs build failure due to h5py


aphearin:
RTD fails to build because of the following problem:

In file included from /tmp/pip-build-7QvtZH/h5py/h5py/defs.c:279:0:
    /tmp/pip-build-7QvtZH/h5py/h5py/api_compat.h:27:18: fatal error: hdf5.h: No such file or directory
     #include "hdf5.h"
                      ^
    compilation terminated.

The complete RTD traceback appears below. The only solution for this has been to simply remove h5py from the install requirements and list it as an optional dependency. However, Halotools makes essential use of h5py, so I'd *really* like to avoid this. 

cc @eteq 

You are using pip version 7.1.0, however version 7.1.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
DEPRECATION: --download-cache has been deprecated and will be removed in the future. Pip now automatically uses and configures its cache.
Requirement already satisfied (use --upgrade to upgrade): numpy in /usr/lib/python2.7/dist-packages (from -r docs/rtd-pip-requirements (line 1))
Requirement already satisfied (use --upgrade to upgrade): matplotlib in /usr/lib/pymodules/python2.7 (from -r docs/rtd-pip-requirements (line 2))
Requirement already satisfied (use --upgrade to upgrade): Cython in /home/docs/checkouts/readthedocs.org/user_builds/halotools/envs/master/lib/python2.7/site-packages (from -r docs/rtd-pip-requirements (line 3))
Requirement already satisfied (use --upgrade to upgrade): astropy-helpers in /home/docs/checkouts/readthedocs.org/user_builds/halotools/envs/master/lib/python2.7/site-packages (from -r docs/rtd-pip-requirements (line 4))
Requirement already satisfied (use --upgrade to upgrade): astropy in /home/docs/checkouts/readthedocs.org/user_builds/halotools/envs/master/lib/python2.7/site-packages (from -r docs/rtd-pip-requirements (line 5))
Requirement already satisfied (use --upgrade to upgrade): scipy in /usr/lib/python2.7/dist-packages (from -r docs/rtd-pip-requirements (line 6))
Collecting h5py (from -r docs/rtd-pip-requirements (line 7))
  Using cached h5py-2.5.0.tar.gz
Requirement already satisfied (use --upgrade to upgrade): python-dateutil in /usr/lib/python2.7/dist-packages (from matplotlib->-r docs/rtd-pip-requirements (line 2))
[...]

PR #298, opened 2016-01-11
Make h5py a travis dependency

eteq:
This implements a change to the tests that I discussed earlier today with @aphearin: it adds `h5py` as a dependency for the travis tests.  Testing `h5py` functionality is generally desired because basically the only reason h5py is not a required dependency is because it won't build on RTD.

This also adds an additional Travis build that's the same as the standard py2.7 build *except* that it does *not* have `h5py` this is here to make sure that an "accidental" dependency isn't injected (e.g., a module-level import of `h5py`), as in that case the RTD builds will start failing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ env:
         # For this package-template, we include examples of Cython modules,
         # so Cython is required for testing. If your package does not include
         # Cython code, you can set CONDA_DEPENDENCIES=''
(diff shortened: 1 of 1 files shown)


aphearin:
Great, thanks @eteq for hopping right on this and making informative in-line comments.

stefco:
Not sure if you found a better fix, but see #154 for officially sanctioned solution

PR #326, opened 2016-01-31
Correct doc for _get_fname() method.

nbeckman:
If I'm understanding things correctly, this comment is backwards.


aphearin:
Thanks, @nbeckman

PR #330, opened 2016-02-02
Mock factory memory improvements

aphearin:
Previously, when populating mocks, the *entire* halo catalog was bound to the mock object. This memory usage was needless. Now only the rows passing the appropriate cuts are stored, and also only the columns required by the model. This should reduce the dominant source of memory usage of the mock factories factory by a few hundred percent. 

CC @mjvakili - this is a pre-requisite to implementing the sample variance feature we have been discussing, but it should immediately improve the memory leak issue you may have been encountering.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ env:
         - SETUP_CMD='test'
         - PIP_DEPENDENCIES=''
         # Here we list the dependencies to be installed that are in conda
-        - CONDA_DEPENDENCIES='cython scipy beautifulsoup4 requests matplotlib h5py'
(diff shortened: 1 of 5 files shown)


rainwoodman:
When we run halotools via the nbodykit wrapper we see the memory usage increase every time mock.populate is rerun. I think it is related to this. I am working on a reproducible test case, but any insights?

aphearin:
@rainwoodman - does the memory usage taper off after a few repetitions, or grow without bound? I think @johannesulf has reported a related issue before?

rainwoodman:
It doesn't taper off. See #917

PR #331, opened 2016-02-02
Fix overflowing text in sphinx sidebar

eteq:
This PR fixes the long-standing problem of long strings in the docs overflowing. Before:
image

After:
image

I'll issue a PR against the astropy-helpers on this so that it goes more widely, so it might be better to hold off on that if this gets merged in there quickly.  But if you want to do a release or just send around cleaner docs before then, then this can just be merged.


eteq:
fix for the helpers is in astropy/astropy-helpers#217

PR #340, opened 2016-02-04
Update helpers to version with fix for overflowing documentation

eteq:
This PR just updates halotools' helpers to include the fix from astropy/astrop-helpers#217

This should do the same thing as #331, but using the updated helpers instead of customizing halotools' css.  So this closes #331 (i.e., #331 should not be merged assuming this works).


eteq:
Oops, looks like #331 got merged before this - just added a commit undoing #331 just so it isn't "double-solved"

PR #401, opened 2016-03-03
Add precompute_RR feature into tpcf

aphearin:
This PR introduces an additional feature to the tpcf function: you can now optionally pass in a set of pre-computed random pairs. This feature is useful for MCMC applications populating multiple simulation volumes, as is currently being done by @mjvakili and @changhoonhahn.


aphearin:
@mjvakili, @changhoonhahn, I'm merging this in to master now.

Issue #335, opened 2016-02-03
include one-halo term integration test


aphearin:
The one-halo term should absolutely always be exactly equal to -1 when evaluated on scales > 30 Mpc. This excellent integration test (suggested by @andrew-zentner) nicely ties together all three main sub-packages of the code: `empirical_models`, `sim_manager` and `mock_observables`.

aphearin:
This Issue is resolved by #404

Issue #341, opened 2016-02-04
Implement satellite position test


aphearin:
Satellite host-centric distances in the HOD should all be less than rvir

aphearin:
This Issue is resolved by #404

PR #404, opened 2016-03-03
Satellite position testing

aphearin:
This PR adds a new non-trivial test of the correctness of satellite positions during mock population. In particular, it is now manually checked that populating a full-scale mock with 'zheng07' into the default simulation results in satellite positions that lie within rvir of their host halo. Previously, this test was only performed at the unit-testing level. This new test is an integration test.


aphearin:
This PR resolves #335 and #341.

PR #410, opened 2016-03-04
Simplify API for users declaring their own halo mass completeness cut

aphearin:
As pointed out by @nickhand, there were several awkward aspects of how the halo completeness cuts were applied during HOD mock generation. This PR addresses this by allowing the user to declare two arguments that may be passed either to the `HodMockFactory` constructor, or equivalently to the `populate_mock` function:

1. Num_ptcl_requirement 
2. halo_mass_column_key

These two arguments together permit the user to make more customizable initial halo completeness cuts with a simpler and more transparent interface. 

This PR also prompted a much-needed syntax cleanup for the `populate_mock` method, which had become very cluttered with nested control flow. 

@nickhand - since you discovered this awkwardness and will be using this feature in short order, would you mind providing some feedback on this implementation? I will wait to merge this PR until you've had a chance to weigh in. 

Also, fyi, I added in a dedicated test to demonstrate that application of this feature is handled correctly: see `TestHodModelFactory.test_Num_ptcl_requirement`

### DIFF
--- a/halotools/empirical_models/factories/hod_mock_factory.py
+++ b/halotools/empirical_models/factories/hod_mock_factory.py
@@ -50,7 +50,8 @@ class HodMockFactory(MockFactory):
 
     """
 
-    def __init__(self, **kwargs):
+    def __init__(self, Num_ptcl_requirement=sim_defaults.Num_ptcl_requirement, 
+        halo_mass_column_key = 'halo_mvir', **kwargs):
         """
         Parameters 
         ----------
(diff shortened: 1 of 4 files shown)


nickhand:
This is exactly what I had in mind for the interface. I think the added transparency is especially useful for new users for the scenario when the completeness cut is applied by accident -- it should now be easy to find the cause via documentation if the low-mass HOD is missing

aphearin:
Great, thanks for the feedback, I'll go ahead and merge this in. Do let me know if you encounter other problems or nuisances.

Issue #293, opened 2015-12-14
create test suite for void statistics


duncandc:
non currently exists

Issue #290, opened 2015-12-14
create test suite for nearest neighbor


duncandc:
non-currently exists

Issue #289, opened 2015-12-14
create test suite for pairwise velocity stats


duncandc:
non currently exixts

PR #414, opened 2016-03-07
Improved testing of mock_observables sub-package

aphearin:
Many new unit-tests added and associated bug-fixes. The functions that are changed are:

1. nearest_neighbors
2. pairwise velocity statistics
3. void statistics

In almost all cases, the bug fixes just addressed edge cases that previously raised uninformative error messages, e.g., if there was no nearest neighbor within a given search radius, the code used to raise a mysterious error message, now it no longer raises an exception. 

The only exception is with the pairwise velocity statistics. In this function, the periodic boundary conditions were not being handled correctly. The source of the bug has to do with the fact that the signs of the velocity was being flipped in an inconsistent way with the fact that the points were being pre-shifted by the pair-counting engine; this bug has been fixed. Note that this bug does not pertain to any other function in mock_observables.

### DIFF
--- a/halotools/mock_observables/nearest_neighbor.py
+++ b/halotools/mock_observables/nearest_neighbor.py
@@ -173,7 +173,9 @@ def _nth_matrix_minimum(m, n, axis=None):
     i,j = m.nonzero()
     d = np.array(m[i,j]).flatten()
     
-    if axis is None:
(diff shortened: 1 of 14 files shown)


aphearin:
This PR resolves #293, #290, #289.

Issue #219, opened 2015-11-19
Cython version bugs


aphearin:
If an incorrect version of cython is installed, the code compiles but the test suite raises totally uninformative errors, pointed out by @elisej. Instead, any call to setup.py should raise an informative exception if cython is out of date. 

CC @eteq

PR #417, opened 2016-03-07
Now requiring cython version 0.23 or greater

aphearin:
Previously, many tests would fail and others would not run if cython.__version__ = 0.19. This PR changes setup.py so that install_requires has cython>0.23.  

CC @eteq


aphearin:
This PR resolves #219 .

PR #424, opened 2016-03-08
Improved docstrings of pair-counters in the double_tree_pairs module

aphearin:
@a-kravtsov - since you have been looking at this code most recently, would mind vetting these documentation improvements and providing feedback on whether this is sufficiently clear, or if further clarification is warranted?


a-kravtsov:
I think it'd be good to explicitly mention that if rbin=0 and sample1=sample2 the code will count number of objects in the sample. As I mentioned, this is not obvious. To me the natural expectation for the number of pairs with zero separation is 0. So mentioning np.diff helps, but if someone uses zero's element of counts, they still may be puzzled.

PR #428, opened 2016-03-10
mock observables bug fixes and API alteration

duncandc:
API of wp altered.

ra_dec_z code bug fix.


duncandc:
@aphearin should alter wp tutorials to use pi_max instead of pi_bins.

PR #442, opened 2016-03-12
Update helpers and bootstrap

eteq:
This updates the bundeled astropy helpers to the latest helpers version, and also copies over the `ah_boostrap.py` from that to the root of halotools.

This should get rid of the deprecation warning from setuptools as well as anything else that might have been fixed in `ah_bootstrap` in the meantime.

### DIFF
--- a/ah_bootstrap.py
+++ b/ah_bootstrap.py
@@ -56,6 +56,12 @@ else:
     _text_type = str
     PY3 = True
 
+
+# What follows are several import statements meant to deal with install-time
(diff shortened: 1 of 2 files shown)


eteq:
@aphearin - you'll want to merge this before doing the release

eteq:
Oh, and it also now includes a tweak to the ``setup.py`` that is in there because an update to the helpers rendered the ``adjust_compiler`` function unnecessary.

PR #457, opened 2016-03-31
Added variable directory argument to helper scripts.

mclaughlin6464:
I've added arguments to store catalogs in a variable location to the startup scripts.


aphearin:
This looks picture perfect, thanks @mclaughlin6464!

PR #505, opened 2016-05-16
Crossmatch

aphearin:
This PR introduces a new user-facing function to the `utils` sub-package, `crossmatch`. This function returns the indices giving the correspondence between elements of an integer array `x` with values that have a match in an integer array `y`. When `x` and `y` have unique entries, this calculation is equivalent to the `join` function in `astropy.table`, but `crossmatch` also covers the case where the array `x` has repeated entries, and the arrays `x` and `y` may only be partially overlapping. This function is useful for transferring properties from one table to another, or also in doing specific kinds of group aggregation calculations in which some pre-computed property of the groups is broadcast to the members (the calculation of host halo mass is a common example). The `crossmatch` function is very fast even for very large arrays, because all the expensive operations are done exclusively using heavily optimized Numpy functions `np.argsort`, `np.unique`, `np.in1d`, `np.searchsorted`, and `np.repeat`. 

Question for @taldcroft - how do *you* do such calculations? Is this functionality redundant with some function in `astropy`? If not, would this be a candidate for the core?

CC @duncandc - nothing for you to do here, just wanted you to be aware that this feature is now in the repo. 

Also tagging @a-kravtsov as we have previously discussed the `np.repeat` function in the context of making mocks; the `crossmatch` source code demonstrates another example of how useful `np.repeat` can be.

### DIFF
--- a/docs/quickstart_and_tutorials/tutorials/catalog_analysis/galcat_analysis/basic_examples/crossmatching_galaxy_catalogs.rst
+++ b/docs/quickstart_and_tutorials/tutorials/catalog_analysis/galcat_analysis/basic_examples/crossmatching_galaxy_catalogs.rst
@@ -1,0 +1,44 @@
+:orphan:
+
+.. _crossmatching_galaxy_catalogs:
+
+****************************************************
+Cross-matching galaxy and halo catalogs
+****************************************************
+
+This tutorial demonstrates how to use the 
+`~halotools.utils.crossmatch` function to create "value-added" versions of your 
+galaxy catalog that are supplemented by additional properties 
+only stored in the halo catalog. 
+
+For a closely related tutorial, see :ref:`crossmatching_halo_catalogs`. 
+
+Example 1: Creating a value-added galaxy catalog from information in a halo catalog
+=====================================================================================
+
+Let's start out by generating a mock galaxy catalog from a (fake) halo catalog:
+
+>>> from halotools.sim_manager import FakeSim
+>>> halocat = FakeSim()
+>>> from halotools.empirical_models import PrebuiltHodModelFactory
+>>> model = PrebuiltHodModelFactory('leauthaud11')
+>>> model.populate_mock(halocat)
+
+The tabular data of mock galaxies is stored in ``model.mock.galaxy_table``, 
+an Astropy `~astropy.table.Table`, and the halo catalog initially passed 
+to the `~halotools.empirical_models.HodMockFactory` is stored in  
+``halocat.halo_table``. All mock galaxy catalogs come with 
+a ``halo_id`` that lets you cross-match the galaxies against the halos they live in 
+using the `~halotools.utils.crossmatch` function. This function returns the indices 
+providing the correspondence between the rows in the ``galaxy_table`` that have 
+matches in the ``halo_table``. 
+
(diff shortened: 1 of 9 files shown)


aphearin:
@taldcroft - Since this function is not table-specific, @eteq suggested a PR into Numpy rather than astropy, which probably makes more sense.

taldcroft:
@aphearin - I haven't had a chance to look in detail, but how is this different from an outer join, which handles non-unique keys?   See e.g. http://astropy.readthedocs.io/en/latest/table/operations.html#identical-key-values

taldcroft:
FWIW I once asked on numpy-dev if they were interested in hstack, vstack and join for structured arrays, and got no response.  Maybe it just got lost in the noise.  The original implementations for astropy were basically just thin wrappers around pure numpy versions.

aphearin:
@taldcroft - I haven't used the outer_join before, but this looks very useful, so I'm glad I asked. I can definitely see using this when I have two tables with repeated entries and I want to combine all their information together. But the use-cases from `crossmatch` are pretty different, and since the calculation I'm doing is simpler, there is a pretty big difference in performance as well. I anticipate doing this inside an intensive MCMC, so I'll need the speed. 

For purposes of discussion, the biggest difference stems from the fact that the join operation requires that the table keys match, which the crossmatch function will not require since it just accepts plain arrays. This can be useful if table2 has information in it that a user might want to port over to table1 for those objects that appear in both tables. I give explicit examples of that use-case in the tutorials I committed with this PR (http://halotools.readthedocs.io/en/latest/quickstart_and_tutorials/tutorials/catalog_analysis/halocat_analysis/crossmatching_halo_catalogs.html); Example 2 shows something I would not know how to do with an outer_join, but maybe you see how to do it? In any event, I find the sort of calculation demonstrated there to be extremely common in both halo and galaxy catalog analysis with Astropy Tables. I can ask around and do a "straw poll" to see whether other people in more observationally-oriented astronomy also do this frequently, but it'd be good to hear your opinion on how you yourself would do the calculation I demonstrated in that Example 2 tutorial. Probably with an aggregation function, I'm guessing?

taldcroft:
@aphearin - probably not enough sleep last night (and not familiar with halotools to grok the examples), but I'm having trouble from the `crossmatch` docstring understanding exactly what it does.  My guess is that it is returning effectively the `idx_1` and `idx_2` values in the table example below (sans masked entries), but I'm not sure:
```
In [16]: t1 = Table({'id': [1, 1, 2, 3, 3, 4, 5], 'idx': [0, 1, 2, 3, 4, 5, 6]})

In [17]: t2 = Table({'id': [0, 1, 3, 5, 8], 'idx': [0, 1, 2, 3, 4], 'info': ['a', 'b', 'c', 'd', 'e']})

In [18]: table.join(t1, t2, keys='id', join_type='outer')
Out[18]: 
<Table masked=True length=9>
  id  idx_1 info idx_2
int64 int64 str1 int64
----- ----- ---- -----
    0    --    a     0
    1     0    b     1
    1     1    b     1
    2     2   --    --
    3     4    c     2
    3     3    c     2
    4     5   --    --
    5     6    d     3
    8    --    e     4
```
For the `crossmatch` docstring it might help to supply a short example like this which fully shows the input and output.

PR #506, opened 2016-05-16
Improved docstring of crossmatch function

aphearin:
The docstring now has clearer language and Astropy Table-based examples, thanks to feedback from @taldcroft and @eteq. 

@taldcroft - the docstring examples are chosen to be simple and do not necessarily demonstrate how the algorithm has a different use-case from `table.join`. To see the difference, the simplest place to look is Example 2 of the halo catalog analysis tutorial.

### DIFF
--- a/halotools/utils/crossmatch.py
+++ b/halotools/utils/crossmatch.py
@@ -6,8 +6,8 @@ import numpy as np
 
 def crossmatch(x, y, skip_bounds_checking = False):
     """
-    Function providing the index-correspondence between the 
-    elements of an integer array x with values that have a match in an integer array y. 
+    Finds where the elements of ``x`` appear in the array ``y``, including repeats. 
+    
     The elements in x may be repeated, but the elements in y must be unique. 
     The arrays x and y may be only partially overlapping. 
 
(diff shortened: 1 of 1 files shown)


taldcroft:
Excellent, this is more understandable.  Your crossmatch example is essentially doing a database "left join" which is broadcasting information from the "right" table onto the "left".  For at least some subset of users just saying "crossmatch gives you the index values to do a left join" will be enough.  And for many that will mean nothing.  :smile:.

Example 2: Transferring host halo properties to their subhalos could be effectively done with the astropy table join like so:
```
>>> halocat = FakeSim()
>>> t = halocat.halo_table
>>> t_host = t[t['halo_hostid'] == t['halo_id']]  # host entries only (is this right?)
>>> t = table.join(t, t_host, keys='halo_id', join_type='left', keys='halo_id',
...                table_names=('', '_host'),
...                unique_col_name='{col_name}{table_name}')
```
This transfers all of the host columns at once with a `_host` suffix, and uses numpy masking to indicate the entries that are not transferred.  You could limit the columns being transferred in the `t_host = ..` statement, and probably the way I wrote it there is repetition but you get the idea.  Using masks can be a blessing or a curse, but in general one cannot count on `0.0` as being a good indicator of missing data versus just a value that happened to be `0`.

I suspect your `crossmatch` will be faster, though the `Table` join is relatively efficient and cythonized.  I think in my original profiling it was within at least a factor of 2 or so of pandas.  Anyway, it is illuminating for me to look at use cases in detail and see if `Table` has the tools to do it.

aphearin:
I'm not sure I follow. Wouldn't this code snippet only transfer the host mass onto hosts? Maybe I just misunderstood, but the point of this calculation is to transfer the mass of the hosts onto their associated subhalos, not just onto the hosts. 

Also, for which version of astropy does table.join support a `unique_col_name` keyword?

taldcroft:
That code *should* transfer the mass of the hosts onto the subhalos.  But it is untested, so maybe I wrote something wrong.  This example transfers `host_property` into the table `t` (which would be the subhalos):
```1
In [2]: t = Table({'id': [1, 2, 3, 4, 5],
   ...:            'host_id': [10, 10, 13, 14, 14],
   ...:            })

In [3]: t_host = Table({'host_id': [10, 11, 12, 13, 14, 15],
   ...:                 'host_property': ['a', 'b', 'c', 'd', 'e', 'f']})

In [4]: table.join(t, t_host, keys='host_id')
Out[4]: 
<Table length=5>
host_id   id  host_property
 int64  int64      str1    
------- ----- -------------
     10     1             a
     10     2             a
     13     3             d
     14     4             e
     14     5             e
```

aphearin:
@taldcroft - looks like I really didn't appreciate how general your `join` function was after all, thanks for clarifying. 

The use-cases are not so different after all, so maybe `crossmatch` shouldn't get merged into the core after all. The main difference is just performance, I guess. I just tried out the `join` example from your last post on a typical halo catalog provided by Halotools. Since your `join` example transferred *all* the columns from t_host onto t, I tried to make the timing testing fair by deleting all other columns besides `hostid` and `halo_mvir`, but keeping the number of rows realistic (~1e6). In order to transfer `halo_mvir`, the `join` function executes in 8s on my laptop, whereas `crossmatch` completes in 600ms. These differences will matter for typical Halotools analyses, so I'll need to keep my `crossmatch` function even though its API directly overlaps with `join`. 

One final thing worth pointing out. My `crossmatch` function improves performance by 3x if I pass in `table[colname].data` rather than `table[colname]`. I was surprised by that since all the loops in `crossmatch` are done in pure Numpy. If this is expected, then disregard this comment, but I thought it was worth mentioning just in case.

taldcroft:
Hmm, bummer that `join` is so much slower, but useful to now have a different implementation that can help with optimizing.  As you have seen there is unfortunately a lot of overhead with `Column` relative to the underlying ndarray.  I think there is room for improvement but it will take some work.

aphearin:
Well, so far my performance needs haven't been hindered by `Column` iteration because you give this nice direct access to the underlying ndarray through the `.data` attribute. So I'm not sure the work in improving performance would be worth the effort, at least not for my needs. I also wouldn't worry too much about the `join` performance. After all, `8s` is awfully fast, and there are probably very few users who have needs to do a `join` inside a truly performance-critical application like an MCMC. Users that do have such needs (like me) are probably able and willing to write their own code that is optimized for their much more restrictive use-case than the `join` function was designed for. I think you're probably just bumping up against the intrinsic trade-off of having a flexible and full-featured function vs. having a heavily optimized function. Now that I know how to use your `join` function better, I'll use `join` rather than `crossmatch` in cases where I want to combine all columns from two overlapping tables, as opposed to just one or two targeted arrays.

Issue #515, opened 2016-05-24
Improve speed of initial load of CachedHaloCatalog


aphearin:
It is annoying how long the initial load into memory takes. This runtime is entirely driven by the number of self-consistency checks that are performed on the halo catalog - simply loading the hdf5 file into memory takes just one second. Some checks may be unnecessary or could be better optimized. 

CC @duncandc

aphearin:
As described in #518 - the only other natural speedup could come from using `crossmatch` instead of `group_member_generator` inside the `broadcast_host_halo_property` function of `value_added_halo_table_functions`. That should be tried before closing this issue.

PR #518, opened 2016-05-25
Halo table read optimization

aphearin:
This PR reduces the runtime it takes to load a CachedHaloTable into memory. Speedup introduced here is ~2x. The only change is to ensure that the table is only read into memory once when performing all the self-consistency checks associated with the `safe_for_cache` test. Further speedups to continue addressing #515.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,7 +43,7 @@ distribute-*.tar.gz
 # Other
 .*.swp
 *~
-
+temporary_testing_script.py 
 *.ipynb
 .ipynb_checkpoints/*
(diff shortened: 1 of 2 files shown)


aphearin:
After running `cProfile` on the loading of a halo table, it's clear what's dominating the runtime: the call to `broadcast_host_halo_property` is sorting the entire table according to `halo_hostid` and then calling `group_member_generator`. It may be faster to use `crossmatch` instead, but in general this pre-sorting is a good idea and will be the bottleneck in table loading.

PR #537, opened 2016-06-01
Delta sigma bugfix

### LINKED ISSUE #523: Bug in units of return value of `delta_sigma`
At a minium, as pointed out by @alexieleauthaud, the units of the return value of the `delta_sigma` function are not correct. One clear contribution to the problem is that the mean matter density is being set to Nptcl/Vbox, whereas this density must be multiplied by the particle mass to have the correct dimensions. However, resolving this bug should instead implement the mean matter density to be determined by the cosmology of the box. See the `empirical_models.phase_space_models.profile_models.profile_helpers.density_threshold` function for an example of how this is done. 

CC @duncandc - I'm self-assigning this one, but I'll ping you to look at this with me next week.

aphearin:
This PR fixes #523. 

The units of the value returned by `delta_sigma` are now Msun/Mpc**2. The mean matter density determines the overall normalization of the signal; it is now computed via `astropy.cosmology`, avoiding shot noise issues that could have been a problem for the previous method. The `delta_sigma` function now requires a `redshift` positional argument. This is used to calculate the appropriate mean matter density for the input `cosmology`.


surhud:
Andrew, the units used for delta_sigma by observers are typically Msun/pc^2 so that the numbers lie between ~0.1 - ~100 on relevant scales and for halos of interest.

I also wanted to suggest clarifying somewhere in the docstrings that this is physical Delta Sigma, or include an optional argument for returning either the physical or the comoving Delta Sigma.

aphearin:
Thanks for pointing that out, @surhud. Let me know whether you think https://github.com/astropy/halotools/pull/537/commits/a15647c500b8a728ada673807b9703ef299f0583 provides the necessary clarification.

surhudm:
Andrew,

I am confused, are the radial bins passed also in physical units? I thought
they were comoving. Then perhaps there is some mixup of comoving and
physical units going on, I do not see the conversion into physical for the
rp values in the code anywhere.

Surhud

aphearin:
No, sorry, good catch. The docstring had it backwards. Fixed by https://github.com/astropy/halotools/pull/537/commits/35a54d60e117c838192bb65c6afa32e76da7f001

surhudm:
Wait it is still not appropriate, as you have to multiply by the comoving
density then, not the physical density. Then everything should be kosher.

I also think you should just write that the units are (h Msun/Mpc^2)
instead of adding the disclaimer that little h=1 later. That is very
confusing as for rp the docs say the units are Mpc/h, making the h explicit.

Sorry about being a pain, this little h bugs creep all the time, and its
better to be careful on the halotools side, as many people may use it as a
blackbox, just reading the docs. Avoiding confusion would be useful.

Surhud

aphearin:
@surhud I've got a question for you. I think it might be useful to support an option for users to be able to pass in precomputed pairs, or a precomputed tpcf. This way, if some user has access to an ultra-optimized pair counter, they only need to pass the results of their own code as an input to `delta_sigma` function, which is then responsible for doing the projection integrals and so forth. 

First of all, do you think this is worth including? Or should I just assume that if `delta_sigma` isn't fast enough, then users need to write their own such function? 

If it's worth including, what would be the most natural way to include it, from a user-facing API perspective? Should users pass in a `tpcf` and `rbins` array? Or would passing in the `counts` result of a pair-counter function make more sense?

surhudm:
I think passing in a tpcf rather than the counts array would make their
life much easier, so they have to just mimic one particular part of the
halotools package.

aphearin:
@surhud - yes, you are right. The comoving normalization correction is made in https://github.com/astropy/halotools/pull/537/commits/5edac430b49fb1e9cb8ead4a0f7d4e1b350243fe.

aphearin:
@surhud - ok, yes, I see your point. It's easier to work directly with the comoving density from the start, as this eliminates the need for a `redshift` argument altogether. See https://github.com/astropy/halotools/pull/537/commits/fdcd2afba14b695bfac662c0cd3635d149502b48.

surhud:
The code looks good now! The xlabel on the x axis of  gg_lensing_tutorial3.png should say h^{-1} Mpc, to be consistent with the label on the y axis, to avoid confusion.

aphearin:
Thanks for all the double-checks for this function, @surhud.

PR #544, opened 2016-06-07
Fixing delta sigma issues

surhud:
Delta sigma and Sigma were off by a factor two. This pull request fixes the trouble Alexie Leauthaud was having in reproducing her Delta Sigma results.

### DIFF
--- a/halotools/mock_observables/two_point_clustering/delta_sigma.py
+++ b/halotools/mock_observables/two_point_clustering/delta_sigma.py
@@ -128,7 +128,7 @@ def delta_sigma(galaxies, particles, rp_bins, pi_max, period,
     :math:`\\Delta\\Sigma` is calculated by first calculating,
 
     .. math::
-        \\Sigma(r_p) = \\bar{\\rho}\\int_0^{\\pi_{\\rm max}} \\left[1+\\xi_{\\rm g,m}(\\sqrt{r_p^2+\\pi^2}) \\right]\\mathrm{d}\\pi
+        \\Sigma(r_p) = 2.0 * \\bar{\\rho}\\int_0^{\\pi_{\\rm max}} \\left[1+\\xi_{\\rm g,m}(\\sqrt{r_p^2+\\pi^2}) \\right]\\mathrm{d}\\pi
 
     and then,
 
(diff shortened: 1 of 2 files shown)


surhud:
@aphearin  I also worry that this kind of bug might also be affecting the wp measurements.

https://github.com/astropy/halotools/blob/master/halotools/mock_observables/two_point_clustering/wp.py

I will check more carefully.

manodeep:
I see the factor of 2 in wp.py. I also compared the output of halotools wp with Corrfunc wp and they agree to about 1e-12. 

(Having just fixed a missing 4/3 factor bug in Corrfunc \xi, these are the kind of bugs that tend to creep by)

surhud:
Ok perfect,  thanks @manodeep, then I will fix the notes in wp.py, because the notes are incorrect!

surhud:
@aphearin I also have a general comment: for unit testing it would be better to use a fixed dataset (say generated using a fixed numpy random seed), and test the output of this function with the output from external codes. It seems the current tests do not cover these kind of errors.

surhud:
Ouch, this volume bug must be moving all your correlation functions down
compared to where they should be, and could affect all your chains you have
run so far. I agree its not fun. But good that you found it.

Surhud

On Tue, Jun 7, 2016 at 1:38 PM, Manodeep Sinha <notifications@github.com>
wrote:

> I see the factor of 2 in wp.py. I also compared the output of halotools wp
> with Corrfunc wp and they agree to about 1e-12.
>
> (Having just fixed a missing 4/3 factor bug in Corrfunc \xi, these are the
> kind of bugs that tend to creep by)
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/astropy/halotools/pull/544#issuecomment-224173805>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAb4jTyUnWtoVjpUrauSbARbty2pZO1zks5qJPXYgaJpZM4Ivhi9>
> .
>



-- 
सुहृद श्रीकांत मोरे.

manodeep:
@surhudm Thankfully no chains were affected - the bug was in \xi and not in wp. But the reason why the bug had escaped the tests was I only checked the number of pairs and not the actual value of \xi. Lesson learned.

surhudm:
I am glad that wp was not affected!

Surhud

aphearin:
Thanks @manodeep and @surhudm. 

Tagging @duncandc so that he is aware of this error. I don't think there are other functions in `mock_observables` for which this factor of 2 is relevant besides `wp`, and for `wp` this factor correctly appears in the implementation (though the `wp` documentation is incorrectly written, and corrected by https://github.com/astropy/halotools/pull/544/commits/062b2bf041054db142f6e42d23967da35d4d4d17). I will scan through the user-facing functions in `mock_observables` to look out for this though. @duncandc if you could do the same that would be helpful.

aphearin:
I am closing this PR as it is superseded by #548, which resolves the Travis failure that is totally unrelated to this bug-fix

PR #557, opened 2016-06-15
Updated citation info

aphearin:
Now providing a link to http://arxiv.org/abs/1606.04106 in the README and also the Citation Info section. This will eventually be updated with the ADS entry and DOI when they become available.


aphearin:
@eteq and @bsipocz - disregard my comments about this PR. Let's chalk up this morning's failures to the mysterious vagaries of CI.

Issue #560, opened 2016-06-25
compiling error when installing v0.2


yymao:
I encountered the following error when installing v0.2

> halotools/mock_observables/isolation_functions/engines/spherical_isolation_engine.cpp:1: error: CPU you selected does not support x86-64 instruction set

The error happened when it was building `marked_cpairs` and `spherical_isolation_engine` (and possibly others). Some Google results seem to suggest something to do with the  compiler option `-march=native` but I am not certain. 

I was trying to install halotools v0.2 with `pip install halotools==0.2` on RHEL 6.8 x86-64 with gcc 4.8.2 (and also tried 4.4.7). On the exactly same system I can install halotools v0.1 without any issues.

PR #561, opened 2016-06-26
Removed -march=native compiler optimization

aphearin:
This PR removes a compiler optimization flag that causes Halotools v0.2 to fail to install on some architectures (RHEL 6.8 x86-64 and also whatever version of OS X used by @nickhand) resolving #560. The source of the failure is still a little mysterious to me, since I successfully installed v0.2 on two different OS X machines, and @duncandc and @eteq both reported successful installations in their OS X platforms. However, both @yymao and @nickhand both reported the same failure. Both have also reported that simply removing this flag resolves the bug. 

The `-march=native` optimization flag was suggested by @manodeep, who reported a ~4x speedup of his pair counters when using this flag. However, on my machine I only get a ~10% speedup when using this flag. Evidently cython is intelligent enough to make up for this in other ways. 

I would like to release Halotools v0.3 **as soon as possible** once this fix is successfully incorporated into the master branch, so once I merge this PR *I would greatly appreciate it if all those tagged in this PR would confirm whether this bug fix resolves the installation problem.* Many thanks in advance.

### DIFF
--- a/halotools/mock_observables/isolation_functions/engines/setup_package.py
+++ b/halotools/mock_observables/isolation_functions/engines/setup_package.py
@@ -15,7 +15,7 @@ def get_extensions():
     include_dirs = ['numpy']
     libraries = []
     language ='c++'
-    extra_compile_args = ['-Ofast', '-march=native']
+    extra_compile_args = ['-Ofast']
 
     extensions = []
     for name, source in zip(names, sources):
--- a/halotools/mock_observables/pair_counters/cpairs/setup_package.py
+++ b/halotools/mock_observables/pair_counters/cpairs/setup_package.py
@@ -16,7 +16,7 @@ def get_extensions():
     include_dirs = ['numpy']
     libraries = []
     language ='c++'
-    extra_compile_args = ['-Ofast', '-march=native']
+    extra_compile_args = ['-Ofast']
 
     extensions = []
     for name, source in zip(names, sources):
--- a/halotools/mock_observables/pair_counters/marked_cpairs/setup_package.py
+++ b/halotools/mock_observables/pair_counters/marked_cpairs/setup_package.py
@@ -16,7 +16,7 @@ def get_extensions():
     include_dirs = ['numpy']
     libraries = []
     language ='c++'
-    extra_compile_args = ['-Ofast', '-march=native']
+    extra_compile_args = ['-Ofast']
 
     extensions = []
     for name, source in zip(names, sources):
--- a/halotools/mock_observables/pairwise_velocities/engines/setup_package.py
+++ b/halotools/mock_observables/pairwise_velocities/engines/setup_package.py
@@ -16,7 +16,7 @@ def get_extensions():
     include_dirs = ['numpy']
     libraries = []
     language ='c++'
-    extra_compile_args = ['-Ofast', '-march=native']
+    extra_compile_args = ['-Ofast']
 
     extensions = []
     for name, source in zip(names, sources):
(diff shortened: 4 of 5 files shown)


duncandc:
My standard "python setup.py build_ext --inplace" still works.

aphearin:
@eteq and @bsipocz - do you know how we can set up Travis so this build failure would have been caught?

yymao:
I can confirm that I can install 7783dd9 on the RHEL 6.8 system with gcc 4.8.2.

Note that the option `-Ofast` requires gcc >= 4.6, and on some older systems (e.g. SLAC machines) the default gcc is < 4.6.

aphearin:
Many thanks for the quick reply, @yymao, and also good point about gcc. 

I find that -Ofast can make a ~30-40% difference, which is worth including if possible. If you or anyone on this chain knows a good way to test within the cython build what version of gcc is used, it may be worth my implementing this so that the `extra_compile_args` list is built dynamically based on the result of such a test. I know that @ngoldbaum has done something similar with yt when assessing whether openMP is available, so tagging him in case he has a suggestion.

aphearin:
Actually, I don't think it's worth it at this stage to bother with automating the compiler flags. I going to handle this as described in #564.  So @ngoldbaum can disregard this chain after all.

nickhand:
for reference, I am running OS X 10.11.5 with gcc 4.9.3 from macports

aphearin:
Wait, @nickhand, sorry, can you clarify? When you said "this fixes the compilation errors", do you mean that you tested the current master branch and it's fixed? Or that you updated your version of gcc?

nickhand:
Apologies for the confusion...I haven't done any updates to my OS or compiler. I am just confirming that this PR fixes the original compilation issues on my machine.

manodeep:
@nickhand I am curious - what was the original install failure? `-march=native` should work on the system you have quoted. (`-march=native` works as advertised on `x86_64` with `gcc>=4.5`; on some `ppc` architecture, `-mtune=native` is the equivalent command. `gcc` docs: https://gcc.gnu.org/onlinedocs/gcc-4.5.4/gcc/i386-and-x86_002d64-Options.html).

nickhand:
@manodeep okay I tried `-mtune=native` and everything compiles when using that flag instead of `-march=native`. It also compiles if I use the built-in `clang` compiler. When using `-march=native` and `gcc`, I have a lot of missing instruction errors, which all look like: 

- no such instruction: `vmovapd %xmm0, %xmm1'
- no such instruction: `vmovsd %xmm0, 8(%rsp)'
- no such instruction: `vcomisd LC161(%rip), %xmm1'

manodeep:
Ahh okay -- that makes sense. What happens is that the `gcc` assembler does not handle the intrinsics correctly. By adding the '-Wa,-q' flag to `gcc`, you can use the `clang` assembler -- should work out of the box if you have `clang` installed via `macports`. If you don't have `macports clang`, then the fix would be to replace `/opt/local/bin/as` with the corresponding assembler for `clang` (a sym-link should work fine). 

You are getting this error because your `gcc` is a real `gcc` rather than `clang` masquerading as `gcc`.

aphearin:
@manodeep - thanks for clarifying the source of the installation problem that @nickhand was having. It is exactly these sorts of obscure machinations that I do not want Halotools users to have to bother with simply to install the code. See #564. 

Thanks to everyone for responding quickly and helping me resolve this installation problem. I'm going to release v0.3 tomorrow now that this is resolved and Halotools installs without problems on red hat, os x, and windows.

PR #566, opened 2016-06-26
Download script bugfix

### LINKED ISSUE #562: bug in download script
When attempting to download higher-redshift catalogs of Bolshoi, there are two problems:

1. The script fails with an error message that would confuse any new user because there are no particles available for the high-z bolshoi snapshots. 

2. The above issue is resolved by throwing the -halos_only flag, but then the script raises the following even less informative exception at the end:

Total runtime to download pre-processed halo catalog = 126.8 seconds

  File "scripts/download_additional_halocat.py", line 177, in <module>
    + new_ptcl_log_entry.fname + "\n\n"
NameError: name 'new_ptcl_log_entry' is not defined

aphearin:
Resolves #562


aphearin:
Discovered bug in recognition of high-redshift multidark catalogs. This should be resolved before merging.

Issue #577, opened 2016-07-01
`modulate_with_cenocc` does not behave as expected


mclaughlin6464:
Currently, in the `Zheng07Sats` model, if `modulate_with_cenocc` is set to true a new `Zheng07Cens` object  is created. This object is the one used to "modulate" the satellite mean, not the one in the larger `HODModelFactory`. So, if the user passes in a different central model or (as in my case) subclasses `Zheng07*` with new parameters, those will be overwritten and not applied correctly. 

I propose that, in lieu of using a boolean variable, the user should pass in the central occupation model they're using to the satellite model if they want to modulate. This shouldn't be challenging, as the user has likely created one anyway, and will guarantee that changes made to the object will be appropriately enacted in the modulation. 

I'm going to make a pull request to implement this change. However, I'm sure this will create issues with people already using the code, so should probably be discussed.

aphearin:
@mclaughlin6464 - good catch on this. I went back and forth on this exact design choice quite a bit before deciding to make the hard-coded choice that any user of Zheng07Sats who wants to use a different central model than Zheng07Cens should just write a different Zheng07Sats. My logic at the time of that decision was that any user using doing modeling this sophisticated would probably be comfortable writing their own satellite model in addition to their own central model. Since at the time of that design choice I was myself the only Halotools user, I just decided to move on and wait until some future power-user came along and actually needed the more generalized behavior. Looks like that time has now arrived.

Here is my worry: if you add this new feature you propose, then the Zheng07Sats model may get intimidatingly complicated to a new user, and this is supposed to be literally the simplest realistic HOD there is. However, I do completely see your point, and I can certainly be convinced that your proposed change would not be such a major design change after all. 

At this point, you are probably by now as much if not more of an expert on the internals of that class as I am, so please do me a favor and think that over some and let me know your opinion. 

I'm going to tag other known users of this class, and also those who have just commented on Zheng07-related default behavior, to see whether they have opinions on this as well. Anyone tagged on the following list should not feel obligated to comment, but if you *do* have an opinion, please speak up within the next week. 
cc @manodeep, @surhudm, @mjvakili, @changhoonhahn, @andrew-zentner

aphearin:
Also @mclaughlin6464 - can you comment on how you discovered this "feature"? Was it through mysterious/unexpected buggy behavior? Or did you know this was going to happen once you decided to switch central occupation models? I ask for purposes of possibly improving documentation if necessary.

mclaughlin6464:
[...]

PR #578, opened 2016-07-01
Resolve issue #577: modulate_with_cenocc changed to cenocc_model

mclaughlin6464:
Fixes issue #577. At the time of this request it is still under debate whether this change is appropriate. However, I believe this is the most natural way to resolve a bug in the way Zheng07Sats handles central modulation. 

Please note I have not run the test modules as my local setup does not support it. I can possible run later, but it would be helpful if someone else could check.

### DIFF
--- a/docs/quickstart_and_tutorials/tutorials/model_building/composing_models/hod_modeling/zheng07_using_cenocc_model_tutorial.rst
+++ b/docs/quickstart_and_tutorials/tutorials/model_building/composing_models/hod_modeling/zheng07_using_cenocc_model_tutorial.rst
@@ -1,0 +1,78 @@
+:orphan:
+
+.. _zheng07_using_cenocc_model_tutorial:
+
+************************************************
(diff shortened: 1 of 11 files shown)


aphearin:
Thanks @mclaughlin6464 - I'll give this a careful review right after the holiday.

mclaughlin6464:
It would appear that the `leauthaud11` model has similar functionality to `zheng07`. It seems we should update that one as well; I'm happy to implemented the same features we have for `zheng07` over the weekend.

mclaughlin6464:
I don't understand the Assembly bias functionality enough to fix this error. I think you'll have to take a look at this @aphearin . No rush.

aphearin:
No problem @mclaughlin6464 - nice work getting it down to just one remaining failure. The HeavisideAssembias class is pretty dense; I didn't see a long-term future in building upon this simple two-population family of toy models, so I didn't bother to make it super extensible. I'll fix the remaining failure next week.

Issue #551, opened 2016-06-09
expose `seed` keyword in HOD mock making for reproducible mocks


nickhand:
For testing purposes, it would be useful if the random seed could be set from the `populate_mock` function of `HodModelFactory` .

It looks like the ``mc_occupation_*`` functions take a `seed` keyword, but I do not see an obvious way of passing in a value when populating mocks (outside of a hack-ish wrapping of those functions).

nickhand:
@aphearin -- I've put together a quick hack that seems to do the trick in my specific case that just wraps the `mc_occupation_centrals` and `mc_occupation_satellites` functions as well as the `mc_pos`, `mc_radial_velocity`, `mc_solid_sphere` functions of the satellites profile, passing in the `seed` keyword. I'm happy to use that in the short term, so whatever timeline you prefer will work for me. Once you have something I'll happily help test out the feature.

aphearin:
Ok, good to know you were able to write a workaround. 

This is sort of a brain burp, but for this purpose there is a particularly handy context manager within Astropy: `from astropy.utils.misc import NumpyRNGContext`. You can see example usage of it all over my test suite (e.g, halotools/mock_observables/two_point_clustering/tests/test_wp.py). Rather than have the Numpy seed be a fully global variable, this context manager sets the seed within its namespace and then resets the value to be whatever it was outside of the context. You should totally ignore this if you don't find it useful, just thought I'd mention it. That's the way I'll modify the source code when the time comes.  I'll try and properly implement this feature soon, keeping this issue open until I do.

aphearin:
This should be resolved by #584. Please re-open if this is not what you had in mind, @nickhand.

PR #584, opened 2016-07-05
Expose random seed keyword argument to mock population

aphearin:
This PR exposes the `seed` keyword to mock population, so all Halotools component and composite models can have entirely deterministic behavior, resolving #551. 

In addition to the above, all appearances of calls to any function in np.random are now protected inside the namespace of the NumpyRNGContext context manager, so that Halotools never modifies the global environment variable controlling the random number seed used by Numpy. In this way, Halotools testing is purely deterministic. Thanks to @eteq for pointing out this very useful Astropy context manager. 

@nickhand - let me know if this provides the behavior you requested.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,7 @@
 0.4 (unreleased)
 ----------------
 
-- No changes yet
+- All models now support an optional ``seed`` keyword argument, allowing for deterministic Monte Carlo realizations of models. As a result of this feature, it is now mandatory that all user-defined models obey a new constraint. Any function appearing in the ``mock_generation_calling_sequence`` must now use the kwargs Python syntax to catch any additional inputs passed to these functions by the MockFactory.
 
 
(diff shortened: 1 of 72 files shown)


aphearin:
As noted in the change log (https://github.com/astropy/halotools/pull/584/commits/f79f4f2aa9844e09b5fdb90609a8195a6b743b1b), with this PR it is now necessary for all component models to obey a new constraint. Any function appearing in the `mock_generation_calling_sequence` must accept Python's **kwargs syntax. This is because the MockFactory now passes in a `seed` keyword argument in addition to a `table` argument. In principle, users could just catch the `seed` argument, but I already anticipate a v0.4 feature in which a `subhalo_table` is passed in for purposes of position assignment, and there will surely be further additional options that will come up as development continues. If users just catch these with kwargs, that would allow for arbitrary extensions of the factory framework.

Issue #583, opened 2016-07-05
Delete all usage of convert_to_ndarray


aphearin:
Should be replaced with np.atleast_1d

PR #585, opened 2016-07-06
Removal of convert_to_ndarray

aphearin:
This PR removes all usage of the utils function `convert_to_ndarray`, which is entirely redundant with the Numpy function `atleast_1d`.


aphearin:
This PR resolves #583.

PR #592, opened 2016-07-14
damn you little h

duncandc:
The prebuilt SMHM models were applying the little h transformations incorrectly.

The models both work with h=0.7 units, but halotools works exclusively in h=1 units.  To keep the parameters the same as quoted in the papers, the internal calculations are done with little h=0.7, and the results are transformed at the end and returned in h=1 units.  This was done "backwords" in the Behroozi model and not applied in the Moster model.    

The hard-coded tests made the same errors, so this did not show up in the tests.  I think this is one of those things that is best caught by using the code, or coding in pairs.  My take-home lesson--whichever way you handled little h... is wrong.  In fact, this correction is probably wrong, and it is getting late, so please take a look!  @aphearin, can you confirm this?


duncandc:
ok, I took a crack at test_leauthaud11_hod.py to fix the test suite failure, but some of the tests for how the occupation number should change when changing the parameters don't seem straight forward.  The results depends on where the <Ncen>(Mhalo) or <Nsat>(Mhalo) is evaluated.  As it is, the evaluation is done right near the cut-off scale, so things are complicated.  I will leave this to @aphearin.

duncandc:
ok, with a patch by @aphearin, this should now be good to go.  I'll wait for tests to pass before merging.

PR #612, opened 2016-07-25
Compile args centralization

aphearin:
This PR moves the definition of the cython `extra_compile_args` argument into a single place used package-wide, making it easier for users to painlessly configure their cython compilation according to the version of gcc they have available to them.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -82,9 +82,9 @@ This will install the master branch of the code that is currently under developm
 	functions by 10-40% (with zero impact on the performance
 	of the mock-making algorithm implemented in `~halotools.empirical_models`).
 	To compile Halotools with these flags thrown,
-	simply add two new elements to the
-	``extra_compiler_args`` list in every source code file
-	named ``setup_package.py``: the string ``'-Ofast'`` and
+	simply change the strings stored in the ``mock_observables_extra_compile_args``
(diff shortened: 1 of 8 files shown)


aphearin:
@eteq and @bsipocz - do you have recommendations for a better choice for how to implement the package-wide global variable `mock_observables_extra_compile_args`?

aphearin:
Well it looks like this approach is flawed after all. Suggestions welcome.

eteq:
@aphearin - hmm... what about using the ``setup.cfg``?  You can use ``configparser.ConfigParser`` to read those files, but the user can override them by editing the file.  

Just to be sure, if I understand right you only need this during setup, right?  In that case you could either do some trickery of having each `setup_package.py` find the ``setup.cfg`` and use that, or read it once in ``setup.py`` (or the lowest-level `setup_package.py` might be safer) and inject it as a variable like `halotools.COMPILER_MO_ARGS` and then have the sub-packages use that.  (Unless that's what the flaw was in your above scheme?)

minor bit: because this is in the setup you might need to work around the fact that `ConfigParser` has a different import in py 2 vs 3.  You can do something like this:
```python
try:
    from ConfigParser import ConfigParser
except ImportError:
    from configparser import ConfigParser
```

PR #618, opened 2016-08-02
Cleaned s_mu_engine of python layer 

aphearin:
Main change is `from libc.math cimport sqrt`


aphearin:
For npts = 1e4, this PR improves the runtime of the `npairs_s_mu` function by many orders of magnitude.

Issue #605, opened 2016-07-22
Moster13SmHm little h


duncandc:
It is ambiguous in the paper what the units are.  It appears by trial and error it should be h=1.

aphearin:
@duncandc - I think you mean Moster10SmHm, yes?

duncandc:
http://halotools.readthedocs.io/en/latest/api/halotools.empirical_models.Moster13SmHm.html#halotools.empirical_models.Moster13SmHm

aphearin:
There was no bug in the implementation. But #620 improved the documentation and in-line comments.

PR #620, opened 2016-08-08
Smhm littleh bugfix

aphearin:
This PR improves the docstrings and in-line comments about how littleh-h is treated in Moster13SmHm implementation.


aphearin:
This PR resolves #605.

PR #621, opened 2016-08-08
Exposed option to modify max_cells_per_dimension in double tree

aphearin:
This PR allows exposes the max_cells_per_dimension to the user to customize performance, which had previously been hidden behind nested function calls by accident. Thanks to @johannesulf for pointing this out. 

@johannesulf - would this modification have been sufficient to keep you from having to rewrite your own tree?


aphearin:
@johannesulf - nevermind, I just remembered that your outer loop was over particles, not cells, so you wouldn't have used this.

PR #623, opened 2016-08-09
Performance improvement of npairs_s_mu_engine.pyx

johannesulf:
I noticed that s_mu_tpcf from from halotools.mock_observables is extremely slow and found that npairs_s_mu_engine.pyx had a very inefficient loop structure. I made small changes to the engine. s_mu_tpcf is now ~3 times faster when using 20 bins in s and mu. Previously, it took ~30 sec for a zheng07 mock (threshold -19) in bolplanck and now takes ~10 sec.

I have checked that the engine (and npairs_s_mu_engine.pyx) give identical results for this particular, non-trivial use case. I did not perform any other tests.

### DIFF
--- a/halotools/mock_observables/pair_counters/cpairs/npairs_s_mu_engine.pyx
+++ b/halotools/mock_observables/pair_counters/cpairs/npairs_s_mu_engine.pyx
@@ -63,6 +63,7 @@ def npairs_s_mu_engine(double_mesh, x1in, y1in, z1in, x2in, y2in, z2in,
     cdef int num_s_bins = len(s_bins)
     cdef int num_mu_bins = len(mu_bins)
     cdef cnp.int64_t[:,:] counts = np.zeros((num_s_bins, num_mu_bins), dtype=np.int64)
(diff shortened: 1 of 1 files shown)


aphearin:
Not sure I follow this at a glance, @johannesulf. Do you know why your changes to the logic only apply to npairs_s_mu, and not, for example, npairs_rp_pi?

johannesulf:
@aphearin Sorry, I made an error with the time it took with the original code. The performance improvement is not nearly as dramatic as originally described. But the new code still makes it faster by a factor of a ~3 for my use case.

The same logic also improves npairs_rp_pi. It does crucially depend on the number of bins, of course. For 40 bins in rp and pi I also get an improvement of order ~2 for tpcf_rp_pi.

aphearin:
Ok, well this is passing all tests so I'll go ahead with the merge. Nice work with the performance enhancement. I'll raise an issue to adapt this for rp_pi purposes.

johannesulf:
Just to update what I did wrong: I originally compared the performance to halotools v0.3. That did not include commit 2e1caba9e2b65980f7922308ad79d8d107e4e520. The speed increase of ~100 came largely from that commit.

Issue #643, opened 2017-01-19
setting redshift on custom model


kilianbreathnach:
In making a custom model with a `HodModelFactory` instance, there doesn't seem to be a simple, clean way of providing a redshift to the constructor out of the box. This would be helpful for my use of halotools and I imagine it would be useful for others too.

The `self.set_model_redshift()` method which is called during `__init__()` searches through `self.model_dictionary` for objects with a `redshift` attribute. So, as of now, the only way to have a non-default redshift initialised in a `HodModelFactory` instance is by providing an instance of some object with a redshift attribute to an arbitrary keyword argument (as a hack). This is obviously built this way to "inherit" the redshift of the `baseline_model_instance` but is a somewhat tedious avenue if you never needed a baseline model, not to mention the circular problem of getting a custom redshift in your baseline model in the first place!

I'd suggest including an optional `redshift` keyword, since many of the `halotools` constructors have this. I attempted this at first and ran into the problem that the keyword is assumed to be a model feature or galaxy type:

> The ``_infer_gal_type_and_feature_name`` method was unable to identify
> the name of your galaxy population from the ``redshift`` key of the model_dictionary.
> If you are modeling a population whose name is neither ``centrals`` nor ``satellites``,
> then you must provide a ``gal_type_list`` keyword argument to 
> the constructor of the HodModelFactory.

I'd be happy to provide a pull request that enables this functionality by avoiding the error in the `model_calling_sequence` and sending a `redshift` kwarg value to the `set_model_redshift()` method to replace any baseline or default values.

As of now the simple solution with `halotools` is obviously:

```
model = HodModelFactory(**kwargs)
model.redshift = redshift
```

aphearin:
@kilianbreathnach - I think I might not quite follow, correct me if I'm wrong, but isn't it the case that if *any* of the component model instances have a `redshift` attribute, then this redshift is guaranteed to get bound to the composite model object created by `HodModelFactory`? At least, I'm sure that was the intention. If I'm right about that, then whenever you build a custom composite model, don't you just need to make sure that one of the components has a `redshift` bound to it? Or do I misunderstand what you're trying to do?

aphearin:
Closing this after merging #644. @kilianbreathnach - please reopen if the issue is not resolved.

PR #644, opened 2017-01-22
Add support for a ``redshift`` keyword argument to the HodModelFactory

aphearin:
This PR resolves the feature request made in #643. 

@kilianbreathnach - I'm done unit-testing this, but would you try it out for the specific use-case that led to #643 and let me know how it goes?

### DIFF
--- a/halotools/empirical_models/factories/hod_model_factory.py
+++ b/halotools/empirical_models/factories/hod_model_factory.py
@@ -120,6 +120,12 @@ class HodModelFactory(ModelFactory):
             Another possible example would be
             gal_type_list = ['centrals', 'satellites', 'orphans'].
 
+        redshift: float, optional
+            Redshift of the model galaxies. Must be compatible with the
+            redshift of all component models, and with the redshift
+            of the snapshot of the simulation used to populate mocks.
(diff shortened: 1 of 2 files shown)


kilianbreathnach:
@aphearin I'm getting the following error:

> AssertionError: You passed an input argument of ``redshift`` = 0.3 to the HodModelFactory
> that is inconsistent with the redshift z = 0.0 defined by the component models

Am I setting something up wrong? This is the code where it breaks:

```
model = HodModelFactory(centrals_occupation=cen_occ,
                        centrals_profile=cen_prof,
                        satellites_occupation=sat_occ,
                        satellites_profile=sat_prof,
                        redshift=0.3)
```

where nothing is done about redshift anywhere previous to that.

aphearin:
So, you have four component models going in. Those are instances so I don't know what their state is from your code. Try inspecting the `redshift` attribute of those instances before passing them to the factory. Depending on which component model you chose, one of them may be implementing the default redshift set in `halotools.sim_manager.sim_defaults.default_redshift`. If that's the case, maybe let me know how I could improve the documentation/warning messages to avoid confusion.

Issue #641, opened 2017-01-12
support for anisotropic boxes in UserSuppliedHaloCatalog?


nickhand:
It seems like it should be possible to make ``Lbox`` a more general 3-vector, instead of just a scalar? If this is something you're open to, I can take a look at a PR.

aphearin:
Fortunately, all of the `mock_observables` functions were written to accept an input `period`, which can either be a scalar or length-3 sequence. And there are actually a relatively small number of places where `Lbox` is even used in either of the `empirical_models` or `sim_manager` sub-packages; I just ran a simple grep on `Lbox` and all appearances are either already compatible with distinct periodicities, or rather trivially changed to be compatible, so that is all good news, since it means that the `UserSuppliedHaloCatalog` requirement of a scalar can be relaxed without too much trouble (the requirement was never there for any particularly good reason). 

I won't be in a position to lead such a PR until the week of 23 January, but I could take a crack at it that week. If you need it sooner, I could certainly review and run tests on a PR sooner than that. On what timescale would you use this feature if it became available?

nickhand:
Great, that's what it seemed from my quick grep on `Lbox` as well. There's no rush on my end to get this implemented -- nbodykit supports anisotropic boxes, so it would be nice to have the compatibility, but nearly all of the use cases for us are isotropic (as are for most people, I would assume). 

I think I should be able to get a PR together that relaxes the scalar requirement in `UserSuppliedHaloCatalog` in the next week or so for review.

aphearin:
Closing this now that #652 has been merged.

PR #652, opened 2017-01-29
adds support for 3-vector Lbox

nickhand:
This PR addresses the feature request #641 

* the `Lbox` attribute of particle and halo catalogs is now stored as a 3-vector
* user can pass in a 3-vector or float


nickhand:
@aphearin - shall I update the change log to note this update as well?  let me know if there is anything else you think this needs

aphearin:
Ahh, right, good point. Yes, please update the changelog. That should be the last package-maintenance task before merging.

aphearin:
I went ahead and ported your feature to `CachedHaloCatalog` for the sake of consistency. This way, Halotools-provided catalogs will the same implementation of the `Lbox` attribute as any user-supplied catalog. If you merge in my PR, I'll incorporate this into master.

PR #673, opened 2017-02-23
Resolved buggy behavior due to incorrect treatment of random number seed

aphearin:
@johannesulf - if you could do the science verification with this PR, that would be much appreciated.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,7 @@
 
 - Added abunmatch sub-package of empirical_models providing conditional abundance matching functionality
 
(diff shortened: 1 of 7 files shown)


aphearin:
For reference - this bug only impacts users who:
1. are populating mocks HOD-style composite models while passing the `seed` keyword to `mock.populate` and/or `populate_mock` methods
2. the composite model makes more than one call to the same `np.random` function - e.g., `np.random.uniform` is called two or more times (so, for example, standard halo occupation models such as `zheng07` and `hearin15` models are unaffected, but `tinker13` includes color modeling is affected)

Issue #665, opened 2017-02-18
Add Conditional Luminosity Function feature


aphearin:
This class has been developed by @johannesulf and will be included in the `v0.5` release.

PR #674, opened 2017-02-25
Cacciato09 Conditional Luminosity Function

johannesulf:
This PR introduces the Cacciato et al. (2009) CLF composite model. Addresses #665.

### DIFF
--- a/docs/function_usage/index.rst
+++ b/docs/function_usage/index.rst
@@ -27,6 +27,14 @@ HOD models
    ../quickstart_and_tutorials/tutorials/model_building/preloaded_models/tinker13_composite_model
    ../quickstart_and_tutorials/tutorials/model_building/preloaded_models/hearin15_composite_model
 
+CLF models
+~~~~~~~~~~
+
+.. toctree::
+   :maxdepth: 1
+
+   ../quickstart_and_tutorials/tutorials/model_building/preloaded_models/cacciato09_composite_model
+
 Subhalo-based models
 ~~~~~~~~~~~~~~~~~~~~~
 
--- a/docs/quickstart_and_tutorials/tutorials/model_building/preloaded_models/cacciato09_composite_model.rst
+++ b/docs/quickstart_and_tutorials/tutorials/model_building/preloaded_models/cacciato09_composite_model.rst
@@ -1,0 +1,148 @@
+.. _cacciato09_composite_model:
+
+*********************************************
+Cacciato et al. (2009) Composite Model
+*********************************************
+
+.. currentmodule:: halotools.empirical_models
+
+This section of the documentation describes the basic behavior of
+the ``cacciato09`` composite CLF model. To see how this composite
+model is built by the `~halotools.empirical_models.PrebuiltHodModelFactory` class,
+see `~halotools.empirical_models.cacciato09_model_dictionary`.
+
+For brevity, we describe this model as a "Conditional Luminosity Function" model. However,
+the associated classes work equally well as a model for the "Conditional Stellar Mass Function".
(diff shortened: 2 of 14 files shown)


aphearin:
@bsipocz - Are y'all having trouble with AppVeyor giving spurious build failures at the moment? All the travis builds with this PR look good but there is a windows build failure that seems to be caused by a cython compilation issue of functions that have been the repo for over a year. 

(Also, FYI, the v0.5 feature freeze begins after this merge is complete).

aphearin:
After lots of experimentation, I've identified the root of the problem causing the AppVeyor failures with this PR. The cython file halotools/empirical_models/occupation_models/engines/cacciato09_sats_mc_prim_galprop_engine.pyx uses the `fmin` and `fmax` functions inside a tightly nested loop, so it is performance-critical that these are cimported functions. In the current implementation:

```
from libc.math cimport fmin, fmax
```

The above import statement causes no problems on any of the Travis builds, but it *does* cause a Windows-specific failure:

```
halotools\empirical_models\occupation_models\engines\cacciato09_sats_mc_prim_galprop_engine.cpp(2663) : 
error C3861: 'fmin': identifier not found

halotools\empirical_models\occupation_models\engines\cacciato09_sats_mc_prim_galprop_engine.cpp(2631) : 
error C3861: 'fmax': identifier not found
```

After some googling around, the reason for this failure appears to be that [fmin and fmax are a C99 features. Microsoft doesn't fully support the C99 standard](http://stackoverflow.com/questions/25537625/when-creating-mex-in-matlab-error-c3861-fmax-identifier-not-found). This problem has also been discussed [here](http://stackoverflow.com/questions/19234452/visual-c-express-fmin-and-fmax-identifier-not-found). 

In fact, it turns out that I myself had encountered this Windows-specific compilation problem in #385. At that time, the "resolution" was to use the Numpy functions `fmin` and `fmax`. However, this does not solve the performance problem since that solution involves going to and from a python layer, which can result in orders of magnitude in performance. 

In the links above, there are several suggestions. One is to use the cpp `algorithm` library; this did not work for me, though it's possible that I simply did not import it correctly:

```
from libcpp.algorithm import fmin, fmax
```

I tried several other ways to import fmin and fmax, including compiling the cython both as c and c++, but in the end, I just wrote fmin and fmax manually, e.g., 

```
# x = fmin(something, something_else)
x = something
if x > something_else:
    x = something_else
```    
This seems to only have percent-level performance implications, and resolves the Windows build problems since fmin and fmax are no longer imported at all.

bsipocz:
@aphearin - I'm sorry not getting back to this earlier. Looking at the conclusions you got to, I'm pretty certain I wouldn't have been a great help anyway.

aphearin:
No problem @bsipocz - the error messages I was seeing on AppVeyor were so obfuscating that I thought something was just wrong on their side, which was why I asked you. I was just unaware of this standard library issue.

surhudm:
Quick question: should the Cacciato 2012 clf model which is just an update
of the 2009 parameters and which fits lensing and clustering be added as a
new module?

aphearin:
@surhudm - if it's just a parameter set update, then this should just be something like a new method that returns the updated values to be stored the param_dict, not a new module

surhudm:
I see, then perhaps this should be called YMB08, since the parameterization
was borrowed from there.

Further, does the module contain information about what the definition of
halo mass used is? Or does the user have to be careful about this? I
believe we used 180 as the mean overdensity in our modeling scheme for this
paper.

Cheers,
Surhud

aphearin:
What I try to do is just make a note in the docstring about which particular mass definition the parameters were fit with. Slightly more aggressive is to raise a warning, but this might be overkill?

surhudm:
I see, I think a docstring would be essential to include where the default
values are set.

Issue #629, opened 2016-08-18
Generalize delta_sigma calculation to support massive neutrinos


aphearin:
As pointed out by @sssaito, the current implementation of the `delta_sigma` function assumes neutrinos are massless. This is reflected in two different ways in the code:
1. The matter density, rho, is computed as mean_rho_comoving = cosmology.Om0_rho_crit0, whereas allowing for massive neutrinos requires mean_rho_comoving = (cosmology.Om0+cosmology.Onu0)_rho_crit0. 
2. Much more significantly, the `sim_manager` is only set up to keep track of a _single_ `ptcl_table` attached to the `halocat`, whereas in simulations with massive neutrinos there are (at least) two species of particles with distinct masses.

PR #696, opened 2017-03-09
Delta sigma from 2d

aphearin:
New implementation of `delta_sigma` function, including API change. Function is more accurate than v0.4 implementation, numerically more stable, faster, and has been generalized to work with simulations that include particles of variable mass, such as hydro or sims with massive neutrinos, resolving #629. 

Special thanks to @surhudm for providing an independent implementation of delta sigma, now included in the unit-testing suite to verify code correctness. 

CC @duncandc @sssaito


surhudm:
@aphearin One quick suggestion for enhancement would be to project along a
given axis rather than z axis alone.

aphearin:
You mean, to help beat down sample variance? Or some other motivation?

surhudm:
Yes, although these are not exactly independent projections...

aphearin:
Exactly, it would be effectively impossible to quote error bars this way.

PR #704, opened 2017-03-16
Fixed a couple of typos in readme

aphearin:
Typos were related to Markdown rendering of strings that include an underscore

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ halocat = CachedHaloCatalog(simname='bolshoi', redshift=0, halo_finder='rockstar
 # populate the catalog with the model
 model.populate_mock(halocat)
 ```
-After calling *populate_mock*, your model will have a *mock* attribute storing your synthetic galaxy population. All Halotools models have a populate_mock method that works in this way, regardless of the features of the model. There are no restrictions on the simulation or halo-finder with which you can use the populate_mock method.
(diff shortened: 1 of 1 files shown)


aphearin:
@bsipocz and @eteq - These CI testing timeouts and MemoryErrors are starting to worry me. They are happening more and more frequently lately, over the past two weeks or so. Sometimes resubmitting the job helps, but for the past day or so this has not resolved itself as CI build failures sometimes do. 

Are either of you seeing Travis and/or AppVeyor problems such as this recently? Or if not, and if the problem is on my side, do you have any experience with common causes for this kind of failure? 

Note that this PR just makes a trivial change to the README, so the fact that I'm getting a failure can only mean that if there is actually a problem in the source code, it must have been introduced in some previous PR.

eteq:
I have never seen a `MemoryError` in this context. Is it possible that it's a legit error in the sense that you're using up all the memory on the test VM? That is, different VMs have different memory limitations​ (I think), so that could explain the intermittent nature. I guess the simple test would be to temporarily skip some of the tests that are failing (and that you know involve a lot of memory)...? If the problem goes away then at least we can try to track down the specific cause knowing it's a memory issue.

That said, the Travis failures are harder to understand... The log just ends. I have seen that before, although the cause is usually something "hard" like a segfault. Perhaps that's also a sign of running out of memory?

eteq:
Oh, another possible debugging approach is to basically mock up what Travis does by installing a VM *locally*, and running the halotools tests. Then just turn down the memory until you start getting errors. If the point that that happens is something reasonable, then that's probably a way to reproduce it.

Issue #707, opened 2017-03-20
crossmatch bug/doc error


duncandc:
The utility function crossmatch(x,y) states that it "finds where the elements of ``x`` appear in the array ``y``, including repeats."

In fact, it finds where the elements of ``x`` appear in the ***sorted*** array ``y``, including repeats."

At the least, the first line of the doc string should indicate this.  I think it would be better if it in fact did what it says currently.  

Any thoughts @aphearin?

duncandc:
I have included some code below demonstrating the use of the code i am envisioning.

First, I create some dummy tables.  
```
import numpy as np

ids = np.arange(0,10).astype('int')

#create mock halo table with id's and some property x
arr_1 = np.recarray((10,), dtype=[('id', int), ('x', float)])
arr_1['id'] = np.random.choice(ids,10,replace=False)
arr_1['x'] = np.random.random(10)

#create mock halo table with id's and some property y
arr_2 = np.recarray((10,), dtype=[('id', int), ('y', float)])
arr_2['id'] = np.random.choice(ids,10,replace=False)
arr_2['y'] = np.random.random(10)
```

Now I can use crossmatch() to match the two arrays into each other using the id's

```
#match the ids in arr1 to arr2
from halotools.utils import crossmatch
inds1, inds2 = crossmatch(arr_1['id'],arr_2['id'])

#matched into sorted versions of each other
result = np.all(arr_1['id'][inds1]==arr_2['id'][inds2])
print("the sorted arrays match eachother: ", result)

#but you cannot put arr1 in the same order as arr2, yet.
result = np.all(arr_1['id'][inds1]==arr_2['id'])
print("arr1 is in the same order as arr2: ", result)
```

In order to put either arr_1 into the order of arr_2 or vice versa, you have to do some numpy gymnastics.  

```
#we need a couple extra steps to put arr1 in the same order as arr2
from halotools.utils import unsorting_indices

s_inds1 = np.argsort(arr_1['id'])
s_inds2 = np.argsort(arr_2['id'])

u_inds1 = unsorting_indices(s_inds1)
u_inds2 = unsorting_indices(s_inds2)

inds1_prime = inds1[u_inds2]
inds2_prime = inds2[u_inds1]
```

now you can put arr_1 into the order of arr_2 and vice versa

```
result = np.all(arr_1['id'][inds1_prime]==arr_2['id'])
print("arr1 is in the same order as arr2: ", result)
```

[...]

PR #744, opened 2017-05-01
updated doc string to indicate the matching is done on the sorted arrays

duncandc:
This is a minimal edit to address issue #707.


aphearin:
I think this is an incorrect description of the function. The input array ``y`` does not need to be sorted in order for this function to work.

duncandc:
I think it is an accurate description, but I agree that it leaves the impression that the user should sort `y`, which of course is not the case.

aphearin:
In any case, as is, this PR damages clarity, not improves it.

Issue #763, opened 2017-06-06
npairs_s_mu performance


aphearin:
As pointed out by @manodeep - the `npairs_s_mu` has needless usage of square roots inside tight cython loops. In exactly the same fashion as with pair counting in Cartesian bins, the engine can be passed `mu_squared_bins` and `s_squared_bins` to avoid this. No changes to the python API are required, this can be handled entirely with a small change to internal variables. 

CC @duncandc @johannesulf @sssaito - I don't know whether performance has been a major problem for y'all, but this will likely speed up `npairs_s_mu` quite substantially.

duncandc:
I haven't noticed s_mu_tpcf being slow, but we should definitely fix this no-brainer issue.

manodeep:
I only took a closer look at the source because `npairs_s_mu` took ~100 seconds on my laptop for the `Mr19` galaxy catalog in `Corrfunc`. For comparison, the equivalent `Corrfunc/theory/DDrppi` takes about 8 seconds on my laptop.

aphearin:
@duncandc - I can handle modifying the internals since it'll just take a second to get a major speedup - but it's uncomfortable to modify these kernels without a reliable brute-force unit-test to calibrate against. Having the `npairs_s_mu` comparison that Shun gave you would be helpful while grinding down the performance.

aphearin:
Actually, @manodeep - since you have recently worked with both codes for the same calculation, would it be easy for you to provide a set of point data, bins, and `npairs_s_mu` counts?

duncandc:
I'll work on the test.  Let me know if someone else is taking over.

manodeep:
@aphearin - that's funny. I was using `halotools` for the test case of known *correct* output :). 

Once I have finished my xi_s_mu (~ month), I will have at least one set of brute-force counts. Happy to send that upstream. I thought @sssaito did have a version of xi(s, mu) derived from the initial (Aspen) version of Corrfunc.

manodeep:
Just created a pull request with the patch. 

On a related note, the inline docs for keep saying `sine` function in the `NOTES` for `mu` in [npairs_s_mu.py](https://github.com/astropy/halotools/blob/master/halotools/mock_observables/pair_counters/npairs_s_mu.py#L63) but the first line of doc says [`cos`](https://github.com/astropy/halotools/blob/master/halotools/mock_observables/pair_counters/npairs_s_mu.py#L60) Why the mis-match?

duncandc:
[...]

PR #768, opened 2017-06-06
Improving npairs(s, mu) performance and testing.

manodeep:
Completely un-tested, un-benchmarked. I am assuming there are existing tests for the npairs(s, mu) pair counters. 

Related to #763

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,7 +1,7 @@
 0.6 (unreleased)
 ----------------
 
-- No changes yet
+- Changed the API for mock_observables.pair_counters.n_pairs_s_mu which now requires ``mu_bins`` to be in the conventional mu=cos(theta_LOS) format instead of mu=sin(theta_LOS).
 
 
 0.5 (2017-05-31)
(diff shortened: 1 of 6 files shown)


duncandc:
Performance is about 10x's worse for me when applying this code.  Are there some compiling flags that we need implement?

aphearin:
Cython's automated compiler flags are sufficiently good that a 10x improvement never occurs by simply adding flags. If you just run `cython -a` on the module, it showed that there were a couple of forgotten declarations that I just added just now.

duncandc:
Excellent.  That did the trick.  It looks to be of order 50% performance increase over the code in the repo right now.   Im going to push my modifications to the pair counters and tests to this PR.

duncandc:
Okay, I plan on merging this with master after it passes the travis build.  I think the brute force pairs.s_mu_npairs comparison test should alleviate any concern that may have been lingering about the reliability of the optimized npairs_s_mu() function.  I have also edited the doc strings of the s_mu pair counters.

duncandc:
@aphearin do you expect some travis builds to fail currently?

duncandc:
okay, note that doc strings with math need to be preceded by an 'r' in order to pass python 3 builds.

Issue #770, opened 2017-06-11
Add `rbins_normalized` feature to pairwise velocity functions


aphearin:
In analogy to the `radial_profile_3d` function, it will be useful to be able to calculate radial velocity dispersions in terms of host-centric distance. 

CC @sssaito - I'll add this feature soon and tag you on the PR.

PR #780, opened 2017-06-13
Add `normalize_rbins_by` feature to radial velocity dispersion calculations

aphearin:
Directly analogous PR to #777, but here for velocity dispersion profiles


aphearin:
Added support for `r/Rvir` calculation, resolving #770.

PR #785, opened 2017-06-22
Addition of Continuous Assembly Bias

mclaughlin6464:
I've implemented my continuous assembly bias model in halotools as a subclass of HeavisideAssembias. I've also made some modifications to how several functions in table_utils work, which makes performing bin conditioned calculations easier. 

The only thing I do not yet have is a suite of tests. I looked at the test suite for Heaviside, and they use existing built-in HOD models. I haven't implemented any of those yet using this model, and I don't know if that's necessarily required for testing.


mclaughlin6464:
A few chances have been made with my model so things won't be merged in for at least awhile. Andrew was being notified every time I made a change, so I'm closing this PR and will open a new one once I'm ready to try to merge.

PR #786, opened 2017-06-22
Installation instruction update

aphearin:
Updated documentation to explicitly walk through how to resolve potential conflicts with previous installations.


aphearin:
@eteq - just want a second opinion on whether the advice I'm giving here is overkill, dangerous, etc.

PR #795, opened 2017-08-14
Added distribution_matching_indices function

aphearin:
The `utils.distribution_matching_indices` function allows you to resample an input distribution (with replacement) so that it matches a desired output distribution. This is useful, for example, when comparing the properties of two galaxy samples with matching stellar mass functions. Visual demonstration shown in the figure below, which is also included in the docs. 

@eteq - do you know whether this function already appears in Astropy?

matched_distributions


eteq:
@aphearin - definitely does not.  I think it definitely could make sense in `astropy.stats`, though, so I'd encourage trying there.  

You *might* get some pushback along the lines of "maybe this should go in `scipy.stats`".  I have a suspicion it wouldn't make it in there (not clear that it's really what `scipy.stats` is about), but worth thinking a bit about. To me this is a fairly astro-common application that I don't see often elsewhere, so I think it's fair to include... But others might disagree.

aphearin:
Ok, sounds good. This should be easy enough to port over to `astropy.stats` since the module is standalone. Thanks for pointing me to the right place.

Issue #797, opened 2017-08-23
Subsampling marks when `max_sample_size` is reached


manodeep:
While computing a [marked correlation function](http://halotools.readthedocs.io/en/latest/api/halotools.mock_observables.marked_tpcf.html), I get the following error from `halotools`:

    HalotoolsError:
    You passed in a 2-D array for `weights2` that
    does not have a consistent shape with `sample2`.
    `sample2` has length 1000000. The input value of `weight_func_id` = 1
    For this value of `weight_func_id`, there should be 1 weights
    per point. The shape of your input `weights2` is (2097152, 1)

I can fix this by simply setting `max_sample_size=bazillion`. 

However, if random down-sampling is applied to the sample, then that **exact** same (random) subsampling should also be applied to the `marks`/`weights`. Good news is that since `halotools` throws the error I got, it will never return any incorrect results.

aphearin:
@manodeep - yeah, I've been using the same workaround myself. I can't really remember why this feature is even in there - maybe a relic of when the code was really slow? Maybe @duncandc remembers. In any case, I use the code with 10^7 points all the time to compute cluster lensing signals for very large mocks, so I also find this annoying. The code runs perfectly fine in this regime, even if not at MCMC speeds. 

Proposed solution: eliminate this feature entirely. Any objections from either @manodeep or @duncandc?

PR #798, opened 2017-08-23
Delete `max_sample_size` feature from `mock_observables`

aphearin:
As pointed out by @manodeep in #797, the `max_sample_size` argument creates a bookkeeping nuisance. I think this feature was originally included at a time when the code was just too slow to use this many points, but that is no longer the case. This PR removes this feature from all `mock_observables` functions. 

Will wait to hear from @duncandc before merging, in case I have forgotten an important reason this feature is here.


duncandc:
There is no good reason for  `max_sample_size`.  Go ahead and merge.

Issue #765, opened 2017-06-06
Improve documentation of Cacciato09Sats


aphearin:
As per discussion with @surhudm and @johannesulf in #674.

johannesulf:
So regarding the naming convention. I think the name Cacciato09 is justified compared to Yang08. The reason is that Yang+08 did not model the normalization of the satellite CLF as a function of halo mass. Instead, it was measured in specific bins of halo mass individually. Additionally, there's a difference in the definition of the normalization of the satellite CLF, $\phi_s^\star$. For a given $\phi_s^\star$ there's a factor of $\log(e)$ difference in the satellite occupation for Yang+08 compared to Cacciato+09. The code follows the convention of Cacciato+09.

johannesulf:
@aphearin @surhudm I addressed this issue by expanding the docstring documentation regarding the published parameters and corresponding halo mass definitions. While I was doing that I found that I used halo_m200b as the definition, but Cacciato+09 used halo_m180b. So that was a tiny error of around 2 - 3% in halo mass. I also added the column halo_m180b to FakeSim so that all tests work with the updated Cacciato09 class. The downside of the new version is that Cacciato09 will not work out of the box with the halotools provided catalogs because they do not contain the halo_m180b column.

PR #807, opened 2017-09-13
Addressed issue #765, changed default halo mass definition for Caccia…

johannesulf:
…to09, added column to FakeSim to work with Cacciato09 by default

### DIFF
--- a/halotools/empirical_models/occupation_models/cacciato09_components.py
+++ b/halotools/empirical_models/occupation_models/cacciato09_components.py
@@ -27,7 +27,7 @@ class Cacciato09Cens(OccupationComponent):
 
     """
 
-    def __init__(self, threshold=10.0, prim_haloprop_key='halo_m200b',
+    def __init__(self, threshold=10.0, prim_haloprop_key='halo_m180b',
                  prim_galprop_key='luminosity', **kwargs):
         r"""
         Parameters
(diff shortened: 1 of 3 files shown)


aphearin:
@johannesulf - have you verified the little-h convention used in halo mass definition is consistent as well? Otherwise, PR looks good.

johannesulf:
@aphearin Cacciato+09 use the usual little-h convention (halo mass in h^{-1} M_\odot) also employed by halotools. So this looks fine.

surhudm:
Hi guys,

Just a quick note: At the time we worked on Cacciato et al 2009, we were
quite unaware of the overdensity problems of FOF halos. So the analytical
model assumes an overdensity of 180b for FOF halos with b=0.2 (whose mass
function and clustering enters the model). In reality the treatment is
inconsistent because we are using halos with higher overdensity (see my
2011 paper with Andrey, as well as cautionary note in Warren 2006) given
the mass function and bias we use, but assuming that they are 180b. A note
should be added somewhere to this effect in the docs. I think this
difference is way larger than the 180b, 200b issue.

I would suggest replacing Cacciato 09 with the more recent Cacciato 2013
model which is based on 200b SO halos, but has a much more consistent
treatment.

Surhud

aphearin:
Well, the models under discussion here have best-fit values that were calibrated using analytical halo models, so their one-to-two halo transitions will differ from the corresponding Halotools prediction even in cases where the same halo mass definition is identical. Users requiring genuinely high-accuracy one-to-two halo transitions will need to refit the parameters of any model that is based on any existing fitting function. Since the pre-built models cannot even in principle reproduce analytical HOD results for this reason, I see little cause to worry too much about this.

PR #812, opened 2017-09-30
Removing `ASTROPY_VERSION=development` from python 2.7.x testing matrix

aphearin:
Development version of astropy appears to have recently started enforcing 3.6 compatibility


aphearin:
Right, yes, I recall the astropy-dev discussion now that you mention it. Thanks for confirming.

Issue #234, opened 2015-11-23
mock_observables: add jackknife wp(rp) function


duncandc:
All the tools exist to do this.

duncandc:
ok, i see this issue already exists!  I'm back on the case.

aphearin:
@duncandc - note that the branch on my fork I made to get you started has been rebased on `master` to accommodate #812. If you already cloned that branch, then rebase on `master` yourself or just reclone.

PR #814, opened 2017-10-11
adding w_p jackknife mock observable

duncandc:
This pull request adds a projected two-point correlation function that calculates jackknife covariance matrices to the mock_observables.two_point_clustering package to address the long-standing issue #234.  

Adding this function required coding up a redshift space jackknife pair counter and the associated Cython engine.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,9 @@
 
 - Added `num_lines_header` optional keyword argument to TabularAsciiReader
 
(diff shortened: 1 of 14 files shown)


aphearin:
@duncandc - is this ready to merge as is? Or were you postponing until you can figure out a good unit-test?

duncandc:
I think this can merge.  I was postponing until we could figure out the jackknife weighting scheme, but let's leave that to a dedicated pull request for both wp and xi if need be.

Issue #244, opened 2015-11-24
covariance matrix function


duncandc:
create a general use covariance matrix calculation function given a matrix of N measurements.  Currently, tpcf_jackknife has its own function, but there is no reason why this could be removed and put somewhere more useful.

duncandc:
this has been included in a new error_estimation_tools module in mock_obs_devel.  soon to be merged with master.

givoltage:
Hi there, I searched in the master and the mock_obs_devel branches and couldn't find this general use function. Was it actually implemented? I'd like to calculate the covariance matrix for all the bin pairs from rp_pi_tpcf. Thanks.

aphearin:
Sorry, @givoltage - it looks like @duncandc dropped his early ambitions to implement this feature. There is currently no code for jackknifing the `rp_pi_tpcf` function. For jackknifing, the only two functions currently supported are `tpcf_jackknife` and `wp_jackknife`. 

I think, but I am not sure, that it would be quite straightforward for `rp_pi_tpcf_jackknife` to be implemented using the same pattern recently used to implement `wp_jackknife`. @duncandc - would this be easy to do, or complicated for reasons I don't foresee?

On a closely related note, in a recent Slack conversation, @andrew-zentner told me that his student Kuan Wang had independently written code to solve the problem for completely generalized two-point function error estimation. @andrew-zentner - since there is now a specific feature request along these lines, do you know how far along Kuan's code is?

duncandc:
This would be trivial to implement.  The workhorse behind wp_jackknife() is npairs_xy_z_jackknife().  The exact same function would also be called in a rp_pi_tpcf_jackknife() function.  I can add this to the list and probably have it done within the week.

On a related note, having a second version of every mock_observable that returns a jackknife covariance matrix may start to get unwieldy.  Maybe Kuan Wang has solved this problem.

PR #822, opened 2017-10-27
adding rp_pi_tpcf_jackknife function to mock_observables 

duncandc:
This pull requests add a jackknife version of the rp_pi_tpcf function.  This feature was requested by @givoltage and discussed in #244.

This function is slightly more complicated than the other jackknife functions.  The observable had to be flattened in order to calculate the covariance matrix.  I would feel more comfortable if someone tries out this function before we fully merge it into master.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,8 @@
 
 - Fixed bug in the normalization of the covariance matrices in the `tpcf_jackknife` function.  See https://github.com/astropy/halotools/issues/815.
 
+- Added `rp_pi_tpcf_jackknife` function.
+
 
 0.5 (2017-05-31)
 ----------------
--- a/halotools/mock_observables/two_point_clustering/__init__.py
+++ b/halotools/mock_observables/two_point_clustering/__init__.py
@@ -11,6 +11,7 @@ from .s_mu_tpcf import s_mu_tpcf
 from .tpcf_multipole import tpcf_multipole
 from .wp import wp
 from .rp_pi_tpcf import rp_pi_tpcf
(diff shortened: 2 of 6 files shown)


aphearin:
@givoltage - have you had a chance to play with this feature for your application yet? We will soon be releasing `v0.6` and were hoping you would be able to provide independent science validation

givoltage:
Thanks much for implementing it. I ran the function successfully without any issue and was going to configure a BAO fit to the output files to see the results. I don't feel quite qualified to comment on the scientific validity of the results now. The only thing I noticed which may or may not be a shortcoming is that output xi_cov is a 2D matrix which takes a little effort to figure out the flattening and indexing convention; a 4D matrix with shape (n_rp_bins-1, n_pi_bins-1, n_rp_bins-1, n_pi_bins-1) would be more straightforward in referencing the cov value to the two bins. But I'm not an expert and maybe there are many reasons why keeping it 2D is desirable.

aphearin:
Thanks for the feedback @givoltage. I think the jackknife function should return a 2D covariance matrix for consistency with the other jackknife functions in `mock_observables`, but I'd be happy to add a few lines to the docs so that users have a simple copy-and-paste one-liner to transform the returned 2D matrix into the 4D matrix that you find more intuitive. Would you be willing to provide the relevant code snippet that performs the flattening and ndarray manipulations that you mentioned? I'm going to go ahead and merge this PR, but I'm happy to alter the docs in a later PR if you could provide the code snippet.

givoltage:
I did not convert the output 2D xi_cov to a 4D array--once I figured out the indexing convention I just took and manipulated the data directly. Thanks.

PR #840, opened 2017-11-29
Add test_installation function 

### LINKED ISSUE #803: simple installation testing
There should be a *much* shorter, but still reliable and robust, way for users to run a test suite to verify that they have a successful installation. And the documentation for this should be immediately discoverable for people who want to quickly pip install and get going. 

@eteq -if you have a suggestion right away, feel free to post here. Otherwise, let's discuss on Slack sometime soon. This has come up repeatedly, since the full test suite can take tens of minutes, which is sure to only become worse over time.

eteq:
@aphearin - I had a look at bf16872 that you added in response to #803. I think I misled you a bit due to my misunderstanding of what goes into the `test()` function's `kwargs` vs `args`.  I *think* this version does the job, though?  Can you give this a try and see if it's what you were thinking? (it's the same as yours except for the last commit changing the `test_installation` function)

If merged, I think this closes #803

### DIFF
--- a/halotools/__init__.py
+++ b/halotools/__init__.py
@@ -6,3 +6,8 @@ and analyzing catalogs of dark matter halos.
 from ._astropy_init import *
 
 from . import custom_exceptions
+
+
+def test_installation(*args, **kwargs):
+    kwargs.setdefault('-m', 'installation_test')
+    return test(*args, **kwargs)
--- a/halotools/__init__.py
+++ b/halotools/__init__.py
@@ -9,5 +9,8 @@ from . import custom_exceptions
 
 
 def test_installation(*args, **kwargs):
-    kwargs.setdefault('-m', 'installation_test')
+    kwargs.setdefault('args', '')
+    if kwargs['args']:
+        kwargs['args'] += ' '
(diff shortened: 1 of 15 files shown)


aphearin:
Hmm, I just tested it and that does not seem to behave as desired. When I run your version, I am now able to do:

```
>>> import halotools
>>> halotools.test_installation()
```
This now successfully runs the test suite, but it runs the *entire* test suite, not just the subset of tests marked by `installation_test` with pytest decorators.

aphearin:
Ok, I just tested this and it behaves exactly as expected - this is perfect, nice work!

eteq:
@aphearin - I think the one travis failure is unrelated, right? If so, I think this is all good.

One could argue this should get a test, but it might not be worth it?  Would have to be a separate travis test that essentially installs and then does ``import halotools;halotools.test_installation()``.  I can try to do that if you like, but it might be better to just merge this and maybe get to it later?

aphearin:
Yes, this test failure is unrelated and mysterious and only happens on some builds. We can ignore it for now. 

I think the test you describe falls pretty squarely in the `overkill` category ;-)

Issue #827, opened 2017-11-14
propagation of keywords when configuring pre-built zheng07


nickhand:
I am trying to configure the pre-built zheng model to use a custom mass column. It's very possible I might be not using this as intended, but I am doing the following:

```python
>>> from halotools.empirical_models import PrebuiltHodModelFactory
>>> model = PrebuiltHodModelFactory('zheng07', prim_haloprop_key='halo_custom_mass', mdef='200c')

>>> for d in model.model_dictionary:
>>>       x = model.model_dictionary[d]
>>>       print(d, getattr(x, 'prim_haloprop_key', None))
centrals_occupation halo_mvir
satellites_occupation halo_custom_mass
centrals_profile None
satellites_profile halo_m200c
```

I think the answer could be that only the "mdef" formatted halo mass columns are supported?
Although it seems to me that ``prim_haloprop_key`` should be getting passed to ``Zheng07Cens`` (the class accepts it as a keyword).

aphearin:
Hmm, it could be that halotools makes a hard requirement for a self-consistent choice of halo boundary and halo mass (halo boundary comes in since satellite positions are generated in `x=r/Rvir`). Marking this as a potential bug to address soon.

nickhand:
I think there are two separate concerns here. Let me explain a little more what we are doing. 

The basic workflow is that the user has a catalog of object positions, velocity, etc, which can be converted into "halos" by specifying a cosmo, redshift, and mass definition. This will add some analytic columns to the catalog based on those parameters, i.e., radius, concentration. Then, the user can specify a halotools model, which will be constructed self-consistently using those parameters, and the halos can be populated with the model. We'd like to be able to pass in those 3 parameters to `PrebuilitHodModelFactory` and have them propagated to the components properly. You can see what we are doing here:

https://github.com/bccp/nbodykit/blob/master/nbodykit/hod.py


Then the problems we ran into:
- `prim_haloprop_key` is not passed to the `Zheng07Cens` class if it is passed as a keyword to  `PrebuiltHodModelFactory` (so it always uses the default mass definition to make `prim_haloprop_key`).
- I don't think there's a good way to support both `mdef` and `prim_haloprop_key`, other than adding explicit checks to see if they disagree (although an exception will be rightly raised eventually). Right now we have the user specify `mdef` and simply assign `prim_haloprop_key` to be the `"halo_m" + mdef` (the default behavior).

nickhand:
This fixes our issue with the Zheng07 model, thanks!

PR #843, opened 2017-11-30
Prim haloprop key propagation

aphearin:
This PR fixes two problems leading to bug report #827:

1. The **zheng07_model_dictionary** function accepts arbitrary kwargs but does not pass them on down to **Zheng07Cens**. No other pre-loaded composite model dictionaries appear to have missed this argument, so this problem appears to have been limited to the **zheng07_model_dictionary** composite model. 
2. The **HodModelFactory** builds a private list called __haloprop_list_ that is the union of each component model's requested properties. Phase space models all must define a quantity _halo_boundary_key_ that will be used to identify the corresponding column of a halo catalog. The _halo_boundary_key_ was not appended to __haloprop_list_. This results in KeyError exceptions when the **HodMockFactory** attempts to populate a halo catalog with a _halo_boundary_key_ column that is actually there. 

Both of these bugs have been fixed, with regression tests for each bug. 

@nickhand - could you report back whether these fixes solve your problem?

### DIFF
--- a/halotools/empirical_models/composite_models/hod_models/zheng07.py
+++ b/halotools/empirical_models/composite_models/hod_models/zheng07.py
@@ -90,7 +90,7 @@ def zheng07_model_dictionary(
     ####################################
     # Build the `occupation` feature
     centrals_occupation = zheng07_components.Zheng07Cens(
-        threshold=threshold, redshift=redshift)
+        threshold=threshold, redshift=redshift, **kwargs)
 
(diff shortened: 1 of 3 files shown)


aphearin:
@nickhand - I'll go ahead and merge this in since this PR does fix bugs that should be fixed. But I'll leave #827 open until you have a chance to test whether this resolves your issue.

PR #845, opened 2017-11-30
Revert unit-testing modules

aphearin:
Reverted files changed by 7b4666f5e7f95c019a6d9b4607fb80cd84555d5d. These were only ever changed to improve testing coverage for edge cases. Doing so led to discovery of several new bugs (which were either resolved with regression tests in separate PRs or are still outstanding). But these changes otherwise did not impact source code performance or behavior, and these are the changes most suspiciously related to the Travis build failures in PRs #843, #842, and #840, which was not resolved by #844.

PR #847, opened 2017-11-30
Removed python=3.5, numpy=1.10 testing configuration

aphearin:
Modified the .travis configuration to drop testing python 3.5 with Numpy v1.10 due to persistent failures that I couldn't track down. The first failure happened in #838, but was resolved after a simple rebuild. #839 didn't have any failures, but then all subsequent PRs failed in this environment configuration, and these failures became persistent. 

In #842, which simply updates the changelog, the entire test suite executes without failure, but then all the doctests located in ``../docs/*.rst`` fail with a mysterious ImportError. 

I attempted to resolve these failures in #845 by rolling back all changes introduced in #838, and in #844 by modifying a couple of specific tests, and in #846 by changing the configuration to Numpy v1.10 with python 3.6 (which is not allowed). I'm basically out of ideas, and so am temporarily abandoning compatibility in python 3.5 with Numpy v1.10 environments, raising #848 as a reminder. 

@bsipocz  and @eteq - if you are aware of any upstream changes that may have caused this, please let me know. I will be ready to release `v0.6` within a week(ish), so please also let me know if you think it's an especially bad idea to drop compatibility with this particular configuration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,7 @@ env:
         - ASTROPY_VERSION=stable
         - SETUP_CMD='test'
         - PIP_DEPENDENCIES=''
+        - PYTEST_VERSION=<3.2
         # Here we list the dependencies to be installed that are in conda
         - CONDA_DEPENDENCIES='cython scipy beautifulsoup4 requests matplotlib h5py'
     matrix:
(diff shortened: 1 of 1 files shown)


aphearin:
Merging this now for the sake of rebasing and incorporating numerous other PRs, some of which are bug-fixes requested by users. Still working with astropy developers on ci-helpers to maintain py3.5 compatibility.

PR #898, opened 2018-03-30
tensor calculations additions and modifications to API

duncandc:
The primary feature added by this pull requests is a reduced inertia tensor calculation.  

In addition, the already existing inertia_tensor_per_object()'s API has been changed to add features and be more consistent with the new function.  

Finally, there is a bug fixed in the matrix_operations_3d.rotation_matrices_from_vectors() and an accompanying test.  Previously, if two identical vectors are passed, a NaN rotation matrix is returned.  I believe the appropriate behavior in that case should be an identity matrix.  

I would like to request a code review by @aphearin and/or anyone else with interest in this calculation.


duncandc:
I found some bugs and I am not yet confident in the calculations.  I will submit another pull request after adding some more tests.

aphearin:
Bugs in existing_inertia_tensor_per_object or new_inertia_tensor_per_object?

Issue #900, opened 2018-04-23
sign_pbc and SubhaloPhaseSpace


johannesulf:
I'm not 100% sure this is a bug, but let's discuss that. SubhaloPhaseSpace uses sign_pbc (or it's internal version _sign_pbc) to calculate relative phase-space positions between subhalos and host halos. sign_pbc throws an assertion error if any phase-space coordinate of a subhalo or host halo meets or exceeds the periodic boundary conditions.

```
assert np.all(x1 < period)
assert np.all(x2 < period)
```

In some cases, this behaviour might be undesired. In my case, some (a handful) of halos have coordinates lying on the periodic boundary conditions. This should in principle not happen but is unavoidable due to rounding errors. (I downloaded a halo catalog from cosmosim.org which has less accuracy than the original ROCKSTAR catalogs.)

Would it make sense to change the above lines?

```
assert np.all(x1 <= period)
assert np.all(x2 <= period)
```

PR #901, opened 2018-04-24
Added support for `x==Lbox` points for SubhaloPhaseSpace model

aphearin:
Thanks to @johannesulf for pointing this out in #900 . 

It might be a good idea if the behavior of a few other Halotools functions such as `enforce_periodicity_of_box` were modified to automatically wrap `x == Lbox` points to `x == 0.` Any thoughts on this @johannesulf?


johannesulf:
I think this might be a good idea. I don't immediately see a downside to this.

PR #904, opened 2018-05-10
Allocate memory before reading chunks

Christopher-Bradshaw:
At the moment, if this is going to fail because of a memory error it will fail right at the end during the concatenate which I think doubles memory (because it assigns a new array and then copies things into it).

With this, we avoid the doubling and if we are going to get an OOM failure we get it much earlier.

I wasn't able to get tests running locally tonight so there is a decent chance I have missed something silly. I can fix up any issues tomorrow, but please let me know if I have missed something and this doesn't make sense.

### DIFF
--- a/halotools/sim_manager/tabular_ascii_reader.py
+++ b/halotools/sim_manager/tabular_ascii_reader.py
@@ -574,7 +574,7 @@ class TabularAsciiReader(object):
         header_length = int(self.header_len())
         print(("Number of rows in detected header = %i \n" % header_length))
 
-        chunklist = []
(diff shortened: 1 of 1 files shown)


Christopher-Bradshaw:
Yeah, what is efficient depends on the cut % and you can't know that before doing it.

My understanding of numpy is that arrays need to be contiguous and so the only way to avoid copying on resize is to use `numpy.ndarray.resize` but that that only works if enough memory at the end of the array is free which is unlikely with the big resizes (100+MB) we want to do here. If there is another way to resize/realloc I don't know about it.

Thanks!

PR #913, opened 2018-06-07
Add option to return indexes in conditional abundance matching

Christopher-Bradshaw:
As discussed in person + slack the primary goal of this is to:
* Add an option to return the index of y2 that was matched (rather than the value in y2) to the CAM code

I made one other minor change - previously, subgrid noise was not being added to the left and right edges. Now it should be.

I've put a number of comments on the PR where I have done something I am not sure about. Let me know what you think!

### DIFF
--- a/halotools/empirical_models/abunmatch/bin_free_cam.py
+++ b/halotools/empirical_models/abunmatch/bin_free_cam.py
@@ -143,11 +143,12 @@ def conditional_abunmatch(x, y, x2, y2, nwin, add_subgrid_noise=True,
         iw += 1
 
 
-    if assume_x_is_sorted:
-        return result
-
     if return_indexes:
-        # I don't think that this is right...
+        # The result indexes point to the location in y2_sorted. Undo that if required
+        result = result if assume_x2_is_sorted else idx_x2_sorted[result]
+        # The result indexes are ordered like y_sorted. Undo that if required
+        result = result if assume_x_is_sorted else result[unsorting_indices(idx_x_sorted)]
         return result
     else:
-        return result[unsorting_indices(idx_x_sorted)]
+        # The result values are ordered like y_sorted, Undo that if required
+        return result if assume_x_is_sorted else result[unsorting_indices(idx_x_sorted)]
--- a/halotools/empirical_models/abunmatch/bin_free_cam.py
+++ b/halotools/empirical_models/abunmatch/bin_free_cam.py
@@ -78,6 +78,8 @@ def conditional_abunmatch(x, y, x2, y2, nwin, add_subgrid_noise=True,
     `~halotools.empirical_models.conditional_abunmatch_bin_based`.
 
     """
+    # add_subgrid_noise = False
+    # return_indexes = True
     if (return_indexes and add_subgrid_noise):
         raise ValueError("Can't add subgrid noise when returning indexes")
 
@@ -140,7 +142,12 @@ def conditional_abunmatch(x, y, x2, y2, nwin, add_subgrid_noise=True,
             result[ix1] = get_value_at_rank(rightmost_sorted_window_y2, rightmost_window_ranks[iw], nwin, int(add_subgrid_noise))
         iw += 1
 
+
     if assume_x_is_sorted:
+        return result
+
+    if return_indexes:
+        # I don't think that this is right...
         return result
     else:
         return result[unsorting_indices(idx_x_sorted)]
--- a/halotools/empirical_models/abunmatch/bin_free_cam.py
+++ b/halotools/empirical_models/abunmatch/bin_free_cam.py
@@ -8,7 +8,25 @@ from .tests.naive_python_cam import sample2_window_indices
 
 
 def conditional_abunmatch(x, y, x2, y2, nwin, add_subgrid_noise=True,
-            assume_x_is_sorted=False, assume_x2_is_sorted=False):
+            assume_x_is_sorted=False, assume_x2_is_sorted=False, return_indexes=False):
+    # Let's say that we want to map SSFR onto a model with Halo + stellar mass
+    # We theorize that SSFR is propto accretion (which we also know from our model)
+
+    # x/y are the model params (e.g. from a sim/theory)
+        # In this case, SM and accretion
+    # x2/y2 are the true distribution (e.g. from observation)
+        # In this case, SM and SSFR
+
+    # Note that the xs are the same thing
+    # But that the Ys contain the things we are modelling on
+
+    # Now, for each x, y (call them xi, yi)
+        # Find a bunch of rows in x that are close to xi. Use their ys to get P(y | xi)
+        # Find the percentile of our y in this distribution
+        # Find a bunch of rows in x2 that are close to xi. Use their ys to get P(y2 | xi)
+        # Find the y2 at the same percentile as our y was in that distribution.
+        # Claim that this y2 is what we expect for this galaxy, given this model
+
     r"""
     Given a set of input points with primary property `x` and secondary property `y`,
     use conditional abundance matching to map new values `ynew` onto the input points
@@ -78,8 +96,11 @@ def conditional_abunmatch(x, y, x2, y2, nwin, add_subgrid_noise=True,
     `~halotools.empirical_models.conditional_abunmatch_bin_based`.
 
     """
+    if (return_indexes and add_subgrid_noise):
+        raise Exception("Can't both return indexes and add noise")
+
     x, y, nwin = _check_xyn_bounds(x, y, nwin)
-    x2, y2, nwin = _check_xyn_bounds(x2, y2, nwin)
+    x2, y2, nwin = _check_xyn_bounds(x2, y2, nwin) # assert these are 1d arrays, nwin is odd
     nhalfwin = int(nwin/2)
     npts1 = len(x)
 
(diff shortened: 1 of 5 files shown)


halotools/empirical_models/abunmatch/bin_free_cam.py
```diff
@@ -3,12 +3,12 @@
...
 from ...utils import unsorting_indices
 from ...utils.conditional_percentile import _check_xyn_bounds, rank_order_function
-from .engines import cython_bin_free_cam_kernel
+from .engines import cython_bin_free_cam_kernel, get_value_at_rank
 from .tests.naive_python_cam import sample2_window_indices
 
 
 def conditional_abunmatch(x, y, x2, y2, nwin, add_subgrid_noise=True,
-            assume_x_is_sorted=False, assume_x2_is_sorted=False):
+            assume_x_is_sorted=False, assume_x2_is_sorted=False, return_indexes=False):
```
Christopher-Bradshaw: I'm not sure what the best api is - another option would be to have two functions:
`conditional_abunmatch` and `conditional_abunmatch_indexes`. I don't know how this is done elsewhere in halotools/how you would like to do it going forward. Your call :)

halotools/empirical_models/abunmatch/engines/bin_free_cam_kernel.pyx
```diff
@@ -159,7 +183,9 @@ def cython_bin_free_cam_kernel(double[:] y1, double[:] y2, int[:] i2_match, int
     cdef int idx_in1, idx_out1, idx_in2, idx_out2
     cdef double value_in1, value_out1, value_in2, value_out2
 
-    cdef double[:] y1_new = np.zeros(npts1, dtype='f8') - 1
+    cdef int[:] y1_new_indexes = np.zeros(npts1, dtype='i4') - 1
```
Christopher-Bradshaw: I tried to define these in if statements:
```
if return_indexes:
    cdef int[:] y1_new_indexes = np.zeros(npts1, dtype='i4') - 1
else:
    cdef double[:] y1_new_values = np.zeros(npts1, dtype='f8') - 1
```
but that wasn't allowed?

Defining both is a bit unnecessary, but I couldn't find a nice way around it

halotools/empirical_models/abunmatch/engines/bin_free_cam_kernel.pyx
```diff
@@ -201,58 +227,47 @@ def cython_bin_free_cam_kernel(double[:] y1, double[:] y2, int[:] i2_match, int
...
         rank1 = correspondence_indx1[nhalfwin]
-        iy2_match = i2_match[iy1]
-
-        #  Stop updating the second window once we reach npts2-nwin/2
-        if iy2_match > iy2_max:
-            iy2_match = iy2_max
+        #  Where to center the second window (making sure we don't slide off the end)
+        iy2_match = min(i2_match[iy1], iy2_max)
 
-        if iy2 > iy2_max:
```
Christopher-Bradshaw: I'm pretty sure that this shouldn't be possible and this set of `if if else` block confused me a bit at first. I think this reads a bit easier now but I'm happy to revert if you disagree.

halotools/empirical_models/abunmatch/engines/bin_free_cam_kernel.pyx
```diff
@@ -201,58 +227,47 @@ def cython_bin_free_cam_kernel(double[:] y1, double[:] y2, int[:] i2_match, int
...
+            index = _find_index(correspondence_indx2, rank1)
+            if index == -1:
+                raise Exception("Index {} not found in correspondence_indx2".format(rank1))
+            y1_new_indexes[iy1] = iy2 + nhalfwin - index
         else:
-            low_rank = rank1 - 1
-            high_rank = rank1 + 1
-            if low_rank < 0:
-                low_rank = 0
-            elif high_rank >= nwin:
```
Christopher-Bradshaw: I think this should just be `if` rather than `elif`? That is what I have in the function (or the same with `min`/`max`)

halotools/empirical_models/abunmatch/tests/test_bin_free_cam.py
```diff
@@ -384,11 +386,13 @@ def test_subgrid_noise1():
     result2 = conditional_abunmatch(x, y, x2, y2, nwin1, add_subgrid_noise=True)
     assert np.allclose(result, result2, atol=0.1)
     assert not np.allclose(result, result2, atol=0.02)
+    assert np.all(result - result2 != 0)
```
Christopher-Bradshaw: This should be enough to assert that we are adding noise to the leftmost/rightmost edges.

halotools/empirical_models/abunmatch/tests/test_bin_free_cam.py
```diff
@@ -503,3 +507,40 @@ def test_initial_sorting4():
...
+
+
+def test_return_indexes():
+    n1, n2 = int(1e2), int(1e2)
+
+    with NumpyRNGContext(fixed_seed):
+        x = np.random.uniform(0, 10, n1)
+        y = np.random.uniform(0, 1, n1)
+
+    with NumpyRNGContext(fixed_seed2):
```
Christopher-Bradshaw: If I used the same seed here as above tests were passing when they shouldn't (before I had correctly reordered/reindexed the indexes). I think that was because we were in the special case of `x == x2` and `y == y2 + c`.

halotools/empirical_models/abunmatch/bin_free_cam.py
aphearin: The only time I prefer creating an entirely new function is in cases where adding in some new feature considerably slows performance; if performance is mission-critical to the original function, I just write a new one. The additional algebraic operations added here are minor enough that I think what you've done here is preferable.

halotools/empirical_models/abunmatch/engines/bin_free_cam_kernel.pyx
aphearin: Actually, I don't know a way around this either. This seems fine.

halotools/empirical_models/abunmatch/engines/bin_free_cam_kernel.pyx
aphearin: Agreed this is a little easier to read, but `min` returns a python object, so it's better practice to use a c-imported min function instead:

`from libc.math cimport fmin as c_min 
`

After looking at the `cython -a` output, in this case I'm pretty sure the cython compiler figures it out anyway and there is no performance hit, so just pointing this out so you are aware of python-cython interactions.

halotools/empirical_models/abunmatch/tests/test_bin_free_cam.py
aphearin: Does this actually ensure that noise was added to the edges? Or just that noise was added *somewhere*?

halotools/empirical_models/abunmatch/tests/test_bin_free_cam.py
aphearin: Right, yes, good catch, sometimes I have to use different seeds for such purposes as well.

halotools/empirical_models/abunmatch/engines/bin_free_cam_kernel.pyx
```diff
@@ -110,11 +110,35 @@ def setup_initial_indices(int iy, int nwin, int npts):
...
+    for i in range(len(arr)):
+        if arr[i] == val:
+            return i
+    return -1
+
+
+@cython.boundscheck(False)
+@cython.nonecheck(False)
+@cython.wraparound(False)
+def get_value_at_rank(double[:] sorted_values, int rank1, int nwin, int add_subgrid_noise):
```
aphearin: Ok, quick implementation comment about this `get_value_at_rank` function.  This function is declared with `def` rather than `cdef`, which means `get_value_at_rank` returns a python object. If you run `cython -a bin_free_cam_kernel.pyx` and inspect the html output, you will see deep yellow on line 270. Click on it and see all the code cython generates to compile this line. To fix this, `get_value_at_rank` needs to be C-declared. See, e.g., the definition of `_bisect_left_kernel`. With this change, this means that `get_value_at_rank` will no longer be accessible by the outside world, but only accessible within cython, so you will not be permitted to add it to `__all__`. 

In general, it's good to get in the habit of always running `cython -a` whenever reviewing cython code. Because cython is a superset of python, the compiler will allow you to write code with avoidable python-layer interactions without complaining, so you just have to be vigilant and always check this manually.

halotools/empirical_models/abunmatch/engines/bin_free_cam_kernel.pyx
```diff
@@ -110,11 +110,35 @@ def setup_initial_indices(int iy, int nwin, int npts):
     return iy, init_iy_low, init_iy_high
 
 
+@cython.boundscheck(False)
+@cython.nonecheck(False)
+@cython.wraparound(False)
+cdef int _find_index(int[:] arr, int val):
```
aphearin: The `_find_index` function implements a blind lookup for a value in the correspondence indices array. The cython is clean (though watch out for the undeclared `index` variable on line 265), but I think the algorithm can be improved by taking advantage of the binary search that is already implemented, no? While the `correspondence_indx2` array is not sorted, the `sorted_cdf_values2` maintains a sorted order. If an original array of indices were defined at the very beginning of the engine, simply with `np.arange(npts2)`, couldn't that array be manipulated alongside `sorted_cdf_values2` to avoid the blind lookup? Naively, it seems this would be faster;  it could be that the performance hit from the additional manipulations wash out the improvement, I don't know, but it seems worth testing since the cythonized binary search machinery is already in place. Let me know whether this makes sense.

halotools/empirical_models/abunmatch/tests/test_bin_free_cam.py
Christopher-Bradshaw: I think it ensures that noise was added to every point

halotools/empirical_models/abunmatch/engines/bin_free_cam_kernel.pyx
Christopher-Bradshaw: I `def`ed it so that it would be accessible to the code that does the left and right edges. 

But maybe a better way would be to have an internal `cdef int _get_value_at_rank` and then an external `def get_value_at_rank` that just wraps this?

Running `cython -a` is definitely a good idea

halotools/empirical_models/abunmatch/engines/bin_free_cam_kernel.pyx
Christopher-Bradshaw: :+1: thanks.

Christopher-Bradshaw: Actually, as these are `ints` I don't think there is a min/max func in `libc.math` for them. (checked https://github.com/cython/cython/blob/master/Cython/Includes/libc/math.pxd)

http://cython.readthedocs.io/en/latest/src/userguide/numpy_tutorial.html suggests that cython has a fast min/max for ints (as you saw in the `cython -a` output).

halotools/empirical_models/abunmatch/engines/bin_free_cam_kernel.pyx
Christopher-Bradshaw: Just to make sure I understand what you are suggesting:
In the same way that we have the `correspondence_indx2` array to keep track of how to unsort the `sorted_cdf_values2`, we can have a another array that keeps track of where things are in the `correspondence_indx2`? That would make this `_find_index` just a lookup in that map.

I **think** that this has roughly the same time complexity - when we update the correspondence indicies (to shift them along) we will need to update this array too and that would be roughly `O(len(nwin))` which is I think what we have now.

But I could well have missed something... I definitely don't fully understand all the internals.

(index `cdef`ed in f05be1c)

aphearin: You may be right, this might not be worth the trouble. After cleaning the cython of the remaining python interactions as discussed elsewhere, could you just do a simple timing test comparing your implementation against what is currently in master? If runtimes are comparable, then it means that further optimization is not worth spending time on.

aphearin: In the timing test, just make sure to use at least ~1e6 points

halotools/empirical_models/abunmatch/engines/bin_free_cam_kernel.pyx
aphearin: `bin_free_cam_kernel.html` looks clean as of this commit. I also just did a simple timing test, comparing your branch to the current version of master, and it looks like there is a nearly negligible performance increase from this feature, so I think this PR is all set once it passes CI.

Christopher-Bradshaw: Sweet, thanks for the comments!

PR #916, opened 2018-06-26
Add idx in cylinders

Christopher-Bradshaw:
Adds an option to the `counts_in_cylinders` function to also get the indexes of the objects in the cylinders. I + a couple of people in our group need this functionality. Performance of just getting the counts is unchanged. Getting the indexes takes ~1.7x the time of counts. This doesn't respond well to parallelization - I think we are mostly memory bound.

The commits are a bit messy but I wanted to leave them as is for the review. I can rebase down later to cleanup if preferred. 

I'll add some comments to lines that I think are interesting, but any feedback is appreciated!


halotools/mock_observables/counts_in_cells/counts_in_cylinders.py
```diff
@@ -191,6 +209,12 @@ def _counts_in_cylinders_process_args(sample1, sample2, proj_search_radius, cyli
     y2 = sample2[:, 1]
     z2 = sample2[:, 2]
 
+    if return_indexes and ((len(x1) > 2e32) or (len(x2) > 2e32)):
```
Christopher-Bradshaw: I don't know when sims will start having > 4 billion objects, but this is already memory intensize and I thought it would be much more common to have someone unable to run this because they were running out of memory vs having someone unable to run this because they have too many objects (and boatloads of ram).

halotools/mock_observables/counts_in_cells/engines/counts_in_cylinders_engine.pyx
```diff
@@ -219,9 +239,24 @@ def counts_in_cylinders_engine(double_mesh, x1in, y1in, z1in, x2in, y2in, z2in,
...
-    idx_unsorted = unsorting_indices(double_mesh.mesh1.idx_sorted)
-    return sorted_counts[idx_unsorted]
-
+                                        if c_return_indexes:
+                                            indexes[current_indexes_cnt, 0] = ifirst1+i
+                                            indexes[current_indexes_cnt, 1] = ifirst2+j
+                                            current_indexes_cnt += 1
+                                            if current_indexes_cnt == current_indexes_len:
+                                                current_indexes_len *= 2
+                                                indexes = np.resize(indexes, (current_indexes_len, 2))
```
Christopher-Bradshaw: I'm not sure if there is a better way to do the resize.

I tried a `Pymem_Realloc` on the pointer to the start `&indexes[0]`of the memoryview and this worked fine, but then something segfaulted later (https://github.com/astropy/halotools/blob/523e7c569ae5f4c65b6aedf2c7148e6593eab4e3/halotools/utils/array_utils.py#L213 the `.astype(int)` piece). I did some reading and as I understand it now you can't (shouldn't?) realloc memoryviews like that?

My though on this is that it will be rare so the python overhead won't be too important? The other options are 1) I have misunderstood something and we can resize memoryviews 2) Use a straight C array.

Happy to do either of them but before digging too deep wanted to get your opinion

halotools/mock_observables/counts_in_cells/tests/test_counts_in_cylinders.py
```diff
@@ -195,7 +195,7 @@ def test_counts_in_cylinders_error_handling():
         approx_cell1_size=0.2, approx_cell2_size=0.2)
 
 
-def test_cic_pbc():
```
Christopher-Bradshaw: Changing these two made it easier to run all `counts_in_cylinders` tests in one go.

Christopher-Bradshaw:
Tests are failing here because of something in the pure python implementation of indexes in cylinders:
```
>       return np.array(indexes, dtype=[("i1", np.uint32), ("i2", np.uint32)])
E       TypeError: data type not understood
```

but have been passing locally. I'll see if I can repro

halotools/mock_observables/counts_in_cells/counts_in_cylinders.py
aphearin: This algorithm would be unusably slow for samples so large. Present-day sims/mocks can exceed these numbers, and Halotools does not have canned functions that deal with that. What is needed is to distribute (buffered) subvolumes across different compute nodes, which is beyond scope for v0.x. So, fine to either delete this line or leave as is, doesn't really matter.

halotools/mock_observables/counts_in_cells/engines/counts_in_cylinders_engine.pyx
aphearin: For practical purposes, since you're not finding the calls to np.resize to be so frequent that it matters, it's fine to just live with np.resize. So comments below do not impact PR acceptance.

Re: segfault. This is perplexing; I can't quite tell what caused this issue since I didn't see what was passed to unsorting_indices with your alternate attempt. Generally, some common causes for this kind of error with python/cython interactions are 
1. mixing up long vs. int ("i8" vs. "i4"), 
2. trying to use a memoryview like a numpy array, e.g., when using the array for fancy indexing. 
Are you sure something simple like that wasn't the problem? 

The reason I ask is that it would be good to know if http://cython.readthedocs.io/en/latest/src/tutorial/memory_allocation.html is giving out-of-date advice on how to dynamically resize memoryviews. If it's not too much trouble to look up, could you point me to where you found this idiom to be deprecated or otherwise not recommended?

halotools/mock_observables/counts_in_cells/engines/counts_in_cylinders_engine.pyx
```diff
@@ -219,9 +239,25 @@ def counts_in_cylinders_engine(double_mesh, x1in, y1in, z1in, x2in, y2in, z2in,
...
+
+    # At this point, we have calculated our pairs on the input arrays *after* sorting
+    # Since the order matters in this calculation, we need to undo the sorting
+    # We also need to reassign these to a non-cdef'ed variables so they can be pickled for pool
+    counts_uns = np.array(counts)[unsorting_indices(double_mesh.mesh1.idx_sorted)]
+    if c_return_indexes:
+        # https://github.com/numpy/numpy/issues/2407 for str("colname")
```
aphearin: Thanks for including details such as this.

halotools/mock_observables/counts_in_cells/tests/test_counts_in_cylinders.py
```diff
@@ -233,3 +233,28 @@ def test_parallel_serial_consistency():
 
     assert result1.shape == result2.shape
 
+@pytest.mark.parametrize("num_threads", [1, 4])
+def test_counts_in_cylinders_with_indexes(num_threads):
```
aphearin: Nicely done. This is compact and very compelling.

halotools/mock_observables/counts_in_cells/engines/counts_in_cylinders_engine.pyx
Christopher-Bradshaw: I think that those docs show how to resize memory blocks manually allocated in C, rather than the memoryview class that I am using here. I think in the example the realloc is OK because that memory is explicitly not managed by python. This is also why it needs to be wrapped it in the `SomeMemory` extension class with a `__dealloc__` method so that it can be freed by python if we return it from the C layer.

The memoryview on the other hand which I allocate here https://github.com/astropy/halotools/pull/916/files#diff-67ca324f5bf87cbe1aa05f366958f7dcR116 is still managed memory and so we can't modify it from the C layer. That messes things up in the Python as things have changed underneath it without it knowing. As you can probably tell I'm not fully sure of the details here but I think it makes sense. If I allocated that as a raw C pointer (e.g. `<double*> PyMem_Malloc(number * sizeof(double))`) then we could realloc as in the example.

The downside of doing that is you need to have the `SomeMemory` extension class that frees memory, and working with C arrays is maybe a bit more tricky/less safe. This didn't seem worth it for something that wasn't hit that much. Also when I originally wrote this I understood this a lot less. Having played with it a bit in the last week I now feel like I understand it a bit better and might be tempted to do it the C way, but I think there are tradeoffs either way.

Issue #917, opened 2018-07-03
Severe memory leak in `populate_mocks()`


rainwoodman:
Run the following test case and we observe the RSS size increase continuously. There is no apparent reason for this to happen reading from the face value of the code. 

```
def test_hearin15():
    """
    """
    model = PrebuiltHodModelFactory('hearin15')
    try:
        halocat = CachedHaloCatalog()
    except:
        halocat = FakeSim()
    import resource
    for i in range(100):
        model.populate_mock(halocat)
        maxrss = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
        print(maxrss)

    assert maxrss < 1000000
```

Result
```
../testenv/lib/python3.6/site-packages/halotools/empirical_models/composite_models/tests/test_preloaded_models.py 169816
185612
199336
216228
230180
230180
230180
241580
257364
276064
286356
291108
302724
315828
330084
345660
359916
...
```

aphearin:
I cannot reproduce your memory leak results using python 2x, @rainwoodman. I have not yet tried python 3.x but will try this soon. 

I can say that the `populate_mock` function was not intended to be used this way. That is a one-time-per-simulation function. After you have called it once, repeated calls on the same simulation should be done using `model.mock.populate()`. Since I cannot reproduce your memory leak test, could you report the results of modifying your function so that `populate_mock` is called prior to the loop, and then `model.mock.populate` is called inside the loop?

It is also worth trying to see whether simply forcing garbage collection resolves the issue, which is what @johannesulf and @surhudm found in the closely related #568.

rainwoodman:
Sorry I was testing this with Python 3.6, actually. This is an updated script, which uses `mock.populate()` the same behavior unless gc is invoked. This indicates there are circular references in the data model. I think if we get rid of the circular references, this can be fixed without requiring triggering gc.

```
def test_hearin15():
    """
    """
    model = PrebuiltHodModelFactory('hearin15')
    halocat = FakeSim()
    import resource
    import gc
    model.populate_mock(halocat)
    for i in range(100):
        model.mock.populate()
        # doesn't matter if model.populate_mock(halocat)
        # gc.collect() # this will cure it.

        maxrss = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
        print(maxrss)

    assert maxrss < 1000000
```

and here is my conda environment list:
```
(base) [yfeng1@waterfall halotools]$ !vim
[...]

PR #918, opened 2018-07-03
Remove reference cycles between MockFactory and ModelFactory.

rainwoodman:
This PR removes the reference cyels between MockFactory and ModelFactory.
I used objgraph to dump the back references in `halotools/empirical_models/composite_models/tests/test_preloaded_models.py::test_hearin15`

Before the PR:
before

After the PR:
after

Unfortunately this does not cure #917. Memory usage still increase after these cycles are cured. There must be other cycles that I haven't found yet.

### DIFF
--- a/halotools/empirical_models/factories/model_factory_template.py
+++ b/halotools/empirical_models/factories/model_factory_template.py
@@ -276,7 +276,9 @@ class ModelFactory(object):
         :ref:`param_dict_mechanism`
         """
 
-        def decorated_func(*args, __param_dict__=self.param_dict, **kwargs):
+        # do not pass self into the scope
+        __param_dict__ = self.param_dict
+        def decorated_func(*args, **kwargs):
 
             # Update the param_dict as necessary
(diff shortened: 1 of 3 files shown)


halotools/empirical_models/factories/model_factory_template.py
```diff
@@ -276,12 +276,12 @@ def update_param_dict_decorator(self, component_model, func_name):
         :ref:`param_dict_mechanism`
         """
 
-        def decorated_func(*args, **kwargs):
+        def decorated_func(*args, __param_dict__=self.param_dict, **kwargs):
```
aphearin: @rainwoodman - this line seems to be causing problems for a few builds. Any idea why? See https://travis-ci.org/astropy/halotools/jobs/399786150.

rainwoodman: It's incompatible with Python 2.7. I pushed a fix.

PR #925, opened 2018-12-23
fix bug in docstring of `wp_jackknife`

syrte:


syrte:
Also see a similar typo https://github.com/astropy/halotools/pull/819

PR #926, opened 2018-12-29
Potential typo

rainwoodman:
Looks like the integration bound shall be c on the denominator?


aphearin:
Thanks for catching this @rainwoodman - this is indeed a typo in the docs.

PR #931, opened 2019-01-23
Fix inconsistent return idx

Christopher-Bradshaw:
In the case where there is only one pair, return indexes would return
```array((1086, 143), dtype=[('i1', '<u4'), ('i2', '<u4')])``` which is not an array and so can't be treated in the same way as the normal output which looks like
```array([(1086, 143), ( ... ),  ], dtype=[('i1', '<u4'), ('i2', '<u4')])```
What I'm pretty sure we want is to be consistent. This changes makes it look lie
```array([(1086, 143)], dtype=[('i1', '<u4'), ('i2', '<u4')])```


aphearin:
Thanks for catching this @Christopher-Bradshaw. I agree this is a more consistent treatment of the n=1 edge case and so this PR improves robustness. The CI failure has nothing to do with your modifications, but is caused by some infrastructure maintenance that I won't be able to get to for a few weeks.

PR #950, opened 2019-10-15
Provide a way to do an autocorrelation in `counts_in_cylinders`

Christopher-Bradshaw:
I don't think there is a way to do an autocorrelation in the `counts_in_cylinders` at the moment. It would be useful for some work I am doing and is a pretty small change.

Let me know if there is anything you want done differently.

Edit: I don't think test failures are this change.


aphearin:
Really nice PR, @Christopher-Bradshaw. The code looks good, and special thanks for proper handling of the unit testing and docstrings.

PR #955, opened 2019-10-28
Mean delta sigma 

aphearin:
Merging in the `mean_delta_sigma` function written by @johannesulf. 

Currently keeping the name `mean_delta_sigma`, although I think this function should outright replace the existing `delta_sigma` function. The question is whether people who use Halotools for lensing calculations would be bothered, so soliciting input from @Christopher-Bradshaw, @dr-guangtou, @alexieleauthaud, and whoever else uses this function.

### DIFF
--- a/halotools/mock_observables/surface_density/__init__.py
+++ b/halotools/mock_observables/surface_density/__init__.py
@@ -1,2 +1,3 @@
 from .delta_sigma import delta_sigma, delta_sigma_from_precomputed_pairs
 from .mass_in_cylinders import total_mass_enclosed_per_cylinder
+from .mean_delta_sigma import mean_delta_sigma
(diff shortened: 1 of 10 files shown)


aphearin:
This PR also introduces a 1-halo and 2-halo decomposition of the lensing signal, using the `mean_delta_sigma_one_two_halo_decomp` function. The new function either accepts halo ID arguments for the input galaxies and particles, or alternatively accepts `halo_radii` for the input galaxies.

Christopher-Bradshaw:
From what we've talked about, it sounds like there is no real downside to the change, just a slightly more accurate estimate? In that case I have no problem with it replacing the old function.

I think there needs to be pretty strong justification to have duplicate functionality else it is confusing for users, and as far as I know there isn't that here.

aphearin:
Hmm, I see your point. The current PR only raises a DeprecationWarning for the old delta_sigma, but maybe you’re right that it’s better for this function to be deleted entirely from the v0.7 release.

Issue #932, opened 2019-01-23
Simple getting started as a contributor docs


Christopher-Bradshaw:
Having not looked at halotools for a year or so, it took a bit of work to remember the code/test workflow that worked with .pyx files. The first time around, this also took me a while. In both cases I worked out that this did what I wanted:
```
python3 setup.py build_ext --inplace # if I changed a .pyx file
cd halotools
pytest [-s -k my_test_seach_string]
```
I'm sure this is very standard and people who work on large, compiled, python projects are probably very used to it. But having this (or a simpler workflow that you use) in a findable place might help new contributors.

aphearin:
The developer docs are definitely lacking! I will be happy to add a new page of the docs dedicated to this. Will get to this before releasing v0.7.

aphearin:
New page added to the docs in #962.

PR #962, opened 2019-11-12
Docs readthrough [WIP]

Christopher-Bradshaw:
This is ready to be looked at.

I've tried to explain why I've done things when it isn't obvious. This is either done with github comments, or comments in the code prefixed with `cbx_` or `cbx_aph`. There are also a couple of open question in the code. Please ping me if anything is unclear.

About 50% of the changes are straight docs changes. The rest are mostly function with unused arguments. In those cases I've remove the args (and made the related docs change!). There are a couple of places where things are a bit more complicated - i.e. the docs claim that something is optional when it is not, or I've noticed something wrong.
 I've tried to only make these more complicated changes when 1) they are a bug, and 2) they are very small/easy to fix. 
I'm sorry if any of these changes are pedantry :)

A crazy idea that I had: There are a ton of duplicated docs (such as the one for `approx_cell1_size` in all the pairfinders. You could define the somewhere and then include them in the docs with a format string. I was briefly excited by this but now think it is a bad idea - too much complexity for minimal gain.

I'm reasonably confident that test failures aren't related to what I've done.

Please feel free to keep/throw away parts of these changes. It's also a pretty spotty lookover - I mainly looked at the parts of HT that I use the most. The codebase is much larger that I thought before I tried to look at all of it :)

### DIFF
--- a/docs/quickstart_and_tutorials/development/getting_started.rst
+++ b/docs/quickstart_and_tutorials/development/getting_started.rst
@@ -47,4 +47,3 @@ You can see the built documentation in ``docs/_build/html/``. The easiest way to
    python3 -m http.server
 
 The docs should then be viewable at ``localhost:8000`` (the port will be logged when you start the server).
-
--- a/docs/quickstart_and_tutorials/development/getting_started.rst
+++ b/docs/quickstart_and_tutorials/development/getting_started.rst
@@ -1,4 +1,4 @@
-.. _getting_started:
+.. _getting_started_developers:
 
 *************************
 Contributing to Halotools
--- a/docs/quickstart_and_tutorials/development/getting_started.rst
+++ b/docs/quickstart_and_tutorials/development/getting_started.rst
@@ -1,0 +1,27 @@
+.. _getting_started:
+
+*************************
+Contributing to Halotools
+*************************
+
+All halotools development happens in the github repository. To contribute, first clone the repo.
+Then install the dependencies listed on the *installation* page.
+
+Halotools contains a compiled component. To compile all `.pyx` files inplace run
+
+```
+python3 setup.py build_ext --inplace
+```
+
+Halotools also has comprehensive unit tests using pytest. From the base of the repository first cd
+into the source tree, then run `pytest`.
+
+```
+cd halotools
+pytest
+```
+
+If you have made a change and only want to run a subset of tests run
+```
+pytest -k name_of_test_to_run
+```
--- a/docs/quickstart_and_tutorials/development/getting_started.rst
+++ b/docs/quickstart_and_tutorials/development/getting_started.rst
@@ -5,23 +5,46 @@ Contributing to Halotools
 *************************
 
 All halotools development happens in the github repository. To contribute, first clone the repo.
-Then install the dependencies listed on the *installation* page.
+Then install the dependencies listed on the :ref:`step_by_step_install` page.
 
-Halotools contains a compiled component. To compile all `.pyx` files inplace run
 
-```
-python3 setup.py build_ext --inplace
-```
+Code
+====
 
-Halotools also has comprehensive unit tests using pytest. From the base of the repository first cd
-into the source tree, then run `pytest`.
+Halotools contains a compiled component. To compile all cython (``.pyx``) files inplace run, ::
 
-```
-cd halotools
-pytest
-```
+   python3 setup.py build_ext --inplace
 
-If you have made a change and only want to run a subset of tests run
-```
-pytest -k name_of_test_to_run
-```
+If you modify a ``.pyx`` file use the same command to recompile it. Subsequent runs will only compile files whose source has changed and so will be much quicker.
+
+Halotools also has comprehensive unit tests and uses pytest. To run all tests, assuming you are in the base of the repository, first change directory into ``halotools`` and then run ``pytest``. ::
+
+   cd halotools
+   pytest
+
+If you have made a change and only want to run a subset of tests run,  ::
+
+   pytest -k tests_matching_this_string_will_run
+
+Run ``pytest --help`` for a full list of options.
+
+
+Docs
+====
+
+First ensure that the halotools package and sphinx are installed. From the base of the repository run, ::
+
+   pip3 install -e .
+   pip3 install sphinx==1.3.1 # see docs/conf.py for the sphinx version
+
+Then build documentation with, ::
+
+   cd docs
+   make html
+
+You can see the built documentation in ``docs/_build/html/``. The easiest way to view it in your browser is to spin up a local server. In the built directory run, ::
+
+   python3 -m http.server
+
+The docs should then be viewable at ``localhost:8000`` (the port will be logged when you start the server).
+
--- a/docs/quickstart_and_tutorials/development/getting_started.rst
+++ b/docs/quickstart_and_tutorials/development/getting_started.rst
@@ -42,7 +42,7 @@ Then build documentation with, ::
    cd docs
    make html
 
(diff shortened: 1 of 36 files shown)


docs/quickstart_and_tutorials/development/getting_started.rst
```diff
@@ -0,0 +1,49 @@
+.. _getting_started:
```
Christopher-Bradshaw: A stab at https://github.com/astropy/halotools/issues/932

halotools/mock_observables/pair_counters/marked_npairs_3d.py
```diff
@@ -66,14 +76,6 @@ def marked_npairs_3d(sample1, sample2, rbins,
         containing the weights used for the weighted pair counts. If this parameter is
         None, the weights are set to np.ones(*(N1,N_weights)*).
 
-    weight_func_id : int, optional
-        weighting function integer ID. Each weighting function requires a specific
-        number of weights per point, *N_weights*.  See the Notes for a description of
-        available weighting functions.
-
-    verbose : Boolean, optional
```
Christopher-Bradshaw: There are a lot of functions that take a `verbose` argument but never use it. I've removed it from most of those.

halotools/mock_observables/pair_counters/marked_npairs_3d.py
```diff
@@ -195,7 +197,7 @@ def _marked_npairs_process_weights(sample1, sample2, weights1, weights2, weight_
     _converted_to_2d_from_1d = False
     # First convert weights1 into a 2-d ndarray
     if weights1 is None:
-        weights1 = np.ones((npts_sample1, 1), dtype=np.float64)
+        weights1 = np.ones(correct_shape1, dtype=np.float64)
```
Christopher-Bradshaw: Previously, if the weight function expected multiple weights, but you didn't pass the weights, this would break. I'm pretty sure that this is what you want. The same change is made in a couple of other places

halotools/mock_observables/pair_counters/npairs_jackknife_3d.py
```diff
@@ -215,19 +219,14 @@ def _npairs_jackknife_3d_process_weights_jtags(sample1, sample2,
...
-    else:
-        jtags1 = np.asarray(jtags1).astype("int")
-        if np.shape(jtags1)[0] != np.shape(sample1)[0]:
-            raise HalotoolsError("jtags1 should have same len as sample1")
+    jtags1 = np.asarray(jtags1).astype("int")
+    if np.shape(jtags1)[0] != np.shape(sample1)[0]:
+        raise HalotoolsError("jtags1 should have same len as sample1")
+
     # Process jtags_2 entry and check for consistency.
-    if jtags2 is None:
```
Christopher-Bradshaw: If this is true, we set them to `0`s but then later error if they are `< 1` and so immediately error. Essentially this means that they are required which is what I have changed the docs + code to reflect.

halotools/mock_observables/pairwise_velocities/engines/velocity_marking_functions.pyx
```diff
@@ -31,12 +31,12 @@ cdef void relative_radial_velocity_weights(cnp.float64_t* w1,
     ----------
     w1 : pointer to an array
         weights array associated with data1.
-        w1[0:2] x,y,z positions
+        w1[0:3] x,y,z positions
```
Christopher-Bradshaw: Total nit...

halotools/mock_observables/pairwise_velocities/los_pvd_vs_rp.py
```diff
@@ -21,57 +21,60 @@
...
 
     do_cross : boolean, optional
-        caclulate the cross-pairwise velocities?
+        calculate the cross-pairwise velocities?
 
     num_threads : int, optional
         number of threads to use in calculation. Default is 1. A string 'max' may be used
         to indicate that the pair counters should use all available cores on the machine.
 
-    seed : int, optional
```
Christopher-Bradshaw: Like verbose, there are a couple of functions that take a seed but never use it.

halotools/mock_observables/two_point_clustering/angular_tpcf.py
```diff
@@ -205,13 +206,14 @@ def area_spherical_cap(chord):
             D2R = N2*(da*rho2)
 
             # calculate the random-random pairs.
+            # cbx: This should be NR(NR-1) if we are doing an autocorr
```
Christopher-Bradshaw: The change/conversation https://github.com/manodeep/Corrfunc/pull/201 explains it in full I think

halotools/mock_observables/two_point_clustering/angular_tpcf.py
```diff
@@ -205,13 +206,14 @@ def area_spherical_cap(chord):
...
             rhor = NR**2/global_area
             RR = (da*rhor)
 
             return D1R, D2R, RR
 
     def pair_counts(sample1, sample2, chord_bins,
-            N_thread, do_auto, do_cross, _sample1_is_sample2):
+            num_threads, do_auto, do_cross, _sample1_is_sample2):
```
Christopher-Bradshaw: `N_thread` is not used inside this function. Instead we use `num_threads` defined in the `angular_tpcf` namespace. It doesn't make any difference to what is actually done, but this argument should either be removed (and we just use `num_threads` in the outer namespace) or we should use the value that is passed. I've done the latter.

halotools/mock_observables/large_scale_density/large_scale_density_spherical_annulus.py
```diff
@@ -105,9 +105,9 @@ def large_scale_density_spherical_annulus(sample, tracers, inner_radius, outer_r
             period, sample_volume, num_threads, approx_cell1_size)
         )
 
-    _ = npairs_per_object_3d(sample, tracers, rbins, period=period,
+    result = npairs_per_object_3d(sample, tracers, rbins, period=period,
```
Christopher-Bradshaw: I think this, and one or two other similar changes, are the only examples of total pedantry. Seeing `_` used confused me just long enough. Sorry...

docs/quickstart_and_tutorials/development/getting_started.rst
aphearin: Wow, thanks for this.

halotools/mock_observables/pair_counters/marked_npairs_3d.py
aphearin: Nice catch!

halotools/mock_observables/pair_counters/npairs_jackknife_3d.py
aphearin: Yup, this is what was intended.

aphearin:
I rebased on ce9bf4ec, which passes on Travis. Everything passes with this PR except for the docs build, so there must have been a breaking change to the docs somewhere. Now trying to track it down.

halotools/mock_observables/large_scale_density/large_scale_density_spherical_annulus.py
aphearin: Ha! Fine.

halotools/mock_observables/two_point_clustering/angular_tpcf.py
aphearin: Hmm, are other autocorrelation function results impacted by this besides angular_tpcf?

aphearin: I went ahead and raised a separate issue #974 since this was the only lingering issue and this is a solid PR that's otherwise good to go.

PR #971, opened 2020-01-07
Include support for empty user-supplied halo catalogs. 

### LINKED ISSUE #936: zheng07_components fails for assembias models.
When using the central modulation feature of Zheng07, if the central model is a subclass of HeavisideAssembias (or a similar model) many calls to mean_occupation will fail. I will submit a pull request that fixes this behavior by defaulting to the baseline_mean_occupation when available.

aphearin:
Resolves #936 

CC @rainwoodman


aphearin:
The Appveyor failure is unrelated and will be resolved elsewhere - Travis passes including the new regression test, so going ahead with the merge.

PR #972, opened 2020-01-07
Fixed bug in treatment of noisy_percentile=-1 edge case. 

### LINKED ISSUE #939: Edge case bug in `noisy_percentile`
I'm finding that when passing in a correlation of -1 to noisy_percentile, i get identical results to passing in 1, but passing in something like -0.9 gives the correct behavior. I think the solution would  be to remove the `abs` in the check condition on line 149, and possibly add another check for -1 that returns percentile[::-1]

aphearin:
Resolves #939. 

cc @mclaughlin6464


aphearin:
Appveyor failure is unrelated - Travis testing passes including the regression test for #939.

Issue #953, opened 2019-10-28
Default halo catalogs should have num_ptcl_per_dim attribute


aphearin:
The following line of code taken from the docstring of `delta_sigma` should not raise an exception:

> total_num_ptcl_in_snapshot = halocat.num_ptcl_per_dim**3

Instead this raises:

> AttributeError: 'CachedHaloCatalog' object has no attribute 'num_ptcl_per_dim'

aphearin:
Likewise, halocat.cosmology also triggers an error

PR #973, opened 2020-01-07
Fixed bug in CachedHaloCatalog treatment of simname

aphearin:
The CachedHaloCatalog was failing to recognize that an input simulation matched a previously cached simulation. The cause of the problem was string-to-bytes comparisons in python 3. This PR resolves the bug by using the `compare_strings_py23_safe` function.


aphearin:
This PR resolves #953.

PR #980, opened 2020-01-08
Remove delta_sigma function from repo

aphearin:
The new function is called mean_delta_sigma


aphearin:
This PR also resolves #981

aphearin:
Users of this function should be aware of the API change for computing lensing with v0.7:

1. The function has changed names from `delta_sigma` --> `mean_delta_sigma`
2. The two arguments `particle_masses` and `downsampling_factor` have been replaced with a single `effective_particle_masses` argument. This new API extends the previous functionality to support particle samples with different downsampling rates for different species. 

cc @johannesulf @alexieleauthaud @Christopher-Bradshaw @villarrealas

PR #987, opened 2020-02-14
Conditional counts in cylinders

AlanPearl:
Wish I could have gotten around to this before the release of 0.7, but I added functionality to the counts_in_cylinders function through keywords 'condition' and 'condition_args'. The only new use I implemented is setting condition="mass_frac" which allows for replicating the star-forming fraction vs neighbor count statistic used in the UniverseMachine paper.

Also, I'm not sure if this was purposefully left out before, but I wrote a line converting the position arrays into 64-bit numpy arrays in the process_args helper function. So now the code works if you pass arrays of np.float32, for example.


halotools/mock_observables/counts_in_cells/engines/counts_in_cylinders_engine.pyx
```diff
@@ -238,7 +268,11 @@ def counts_in_cylinders_engine(
...
 
                                     if (dxy_sq < rp_max_squaredtmp) & (dz_sq < pi_max_squaredtmp):
-                                        counts[ifirst1+i] += 1
+
+                                        i_original = idx_sorted1[ifirst1+i]
+                                        j_original = idx_sorted2[ifirst2+j]
+                                        if cond.func(i_original, j_original):
+                                            counts[ifirst1+i] += 1
 
                                         if c_return_indexes:
```
aphearin: Does this line also need to be modified in order to get the condition function to work properly with the return_indexes feature?

aphearin:
This looks like a nice addition to the functionality of the CIC function, thanks @AlanPearl! 

I can imagine extensions along the lines of the "marking function" framework that we built for the marked correlation functions, which would support a range of different functions in addition to m2/m1-based inequalities. However, I think the equivalent of that pattern could easily work with the way you've written this, so further generalizations could just be built on top of what you've done and handled on an as-needed basis.

So this PR looks like it's close to being ready to merge since the only CI test failures are due to documentation build problems, which I will try and resolve - no need for you to worry about that. But one thing that would be really helpful is a unit-test or two validating the source code. One good test would be to set up sample1 and sample2 particles such that the mass ratios of pairs are always the same, and enforce that the normal CIC result agrees exactly with the marked-CIC result. I can help with this, too, but since you've been actually using this function to do science then maybe this wouldn't be too much additional work?

AlanPearl:
I nested the lines keeping track of the indexes so that they are only run if the condition is true. Also, let me know if the unit tests I made are sufficient. I made it a simple case where sample1 = sample2 and the points are spaced far enough that only self-pairs are counted. But it tests two sides of an edge case of whether or not to allow the exact mass ratio of m2/m1 = 0.5.

aphearin:
Thanks for taking the time to include some unit-testing infrastructure @AlanPearl. Your False and True tests look similar to other testing setups I've used in the rest of mock_observables, so these are reassuring. I think it should be pretty easy for me to riff on what you've set up here and adapt some of the other unit tests I've written into the pattern you've set up. I'm happy to lend that elbow-grease now that you've laid the groundwork (to mix metaphors); I should be able to get to this by the end of the week if you can wait that long until this is merged into master.

PR #1001, opened 2020-11-23
fixed bug for weighted nearest integer distributions

johannesulf:
Currently, the function `_weighted_nearest_integer` function returns floats when it should return integers. This leads to crashes if one wants to populate a halo catalog with an HOD model that has non-Poisson numbers for satellites. This PR fixes that.


aphearin:
Whoops, sorry for the bug, thanks for the fix!

Issue #993, opened 2020-08-24
Relax requirement for consistency between halo mass and radius


aphearin:
Thanks to @jmsull and @rainwoodman for pointing out the problem this is causing in https://github.com/bccp/nbodykit/issues/634.

aphearin:
@jmsull  - would you mind letting me know whether #995 resolves the issue for you?

jmsull:
@aphearin It seems like it does not? I installed halotools from source with the new changes and reran the example in the nbodykit issue and got the same error (first screenshot). I tried running the full halotools test suite (using` halotools.test()`) and the test for #995 seems to pass (second screenshot).
<img width="1115" alt="Screen Shot 2020-09-26 at 02 30 40" src="https://user-images.githubusercontent.com/50674574/94344068-ca51a580-ffea-11ea-9ce7-d1eb6e79b5e8.png">

<img width="1117" alt="Screen Shot 2020-09-26 at 02 35 26" src="https://user-images.githubusercontent.com/50674574/94344643-b60fa780-ffee-11ea-8c5a-0bd707e1786b.png">

jmsull:
@aphearin Sorry it does not seem to resolve the issue - without changing anything in the example code I get the same error as before (screenshot 1). I tried running the test suite and the new test in #995 seems to pass (screenshot 2). Do I need to explicitly pass `halo_boundary_key` somehow?

<img width="1115" alt="Screen Shot 2020-09-26 at 02 30 40" src="https://user-images.githubusercontent.com/50674574/94345334-23bdd280-fff3-11ea-9920-ce7a0200d2b5.png">

<img width="1117" alt="Screen Shot 2020-09-26 at 02 35 26" src="https://user-images.githubusercontent.com/50674574/94345338-27515980-fff3-11ea-86bc-0a4139b3e843.png">

aphearin:
@jmsull - yes, sorry, I forgot to mention that the fix requires you to pass in a halo_boundary_key.

jmsull:
@aphearin Sorry for the extreme lag on this. This did not end up working for me at the time. I tried passing halo_boundary_key kw through nbodykit (and through halotools alone) to no avail.
[...]

PR #1008, opened 2021-02-19
WIP PR resolving #993

aphearin:
Thanks to @jmsull for helping track this down. This WIP PR is getting close but is not quite there.


aphearin:
This will resolve #993 once merged.

PR #1018, opened 2021-08-03
Fixed bug in mock_observables/two_point_clustering/tpcf_jackknife.py

nehapjoshi:
Source code for tpcf_jackknife.py raised an error for cross-correlations where do_auto = False was set.  New edits follow the format of wp_jackknife.py to allow cross-correlations to run without issue, and give correct outputs.

PR #1021, opened 2021-08-04
Intrinsic Alignment Mock Observables

aphearin:
This PR brings in code led by @duncandc implementing synthetic measurements of two-point clustering marked by intrinsic alignments. The source code is pulled in from a branch used by @nehapjoshi. Before merging in this functionality, I'd like to make sure that the code is compatible with what @nvanalfen @jablazek @c-d-leonard and @EiffL may have been using, so that I don't inadvertently break their code. Once this is merged in, I think it will be time to consider a feature-freeze, to resolve outstanding issues and existing PRs, and release v0.8.

Note that this PR obviates #1018.


nvanalfen:
It looks like this should all still be compatible with the code I'm using.

PR #1026, opened 2021-10-08
Added numpy installation requirement

aphearin:
Modified setup.py to reflect numpy installation requirement. Thanks to @lgarrison for catching this when reporting #1025.


aphearin:
Hmm, looks like the tests began silently failing at some point, since the most recent commit makes no changes to the repo but both workflows still fail

PR #1028, opened 2021-10-14
Zu hod bugfix

aphearin:
Fixed bug in the central occupation function of the Zu & Mandelbaum 2016 HOD. Was previously using natural exponentiation instead of base-10. It would be great if @AlanPearl could confirm that his SV tests pointed to this same specific fix before this gets merged into master.


AlanPearl:
Hmm, looks like you changed the wrong exp^() to 10^(). I think the one in ZuMandelbaum15Sats.mean_occupation seemed to be working correctly as it was. The line that needs changing to base-10 is here: https://github.com/astropy/halotools/blob/dfc817a9253268d6449bb1f91b6b2a81fcc6ac7f/halotools/empirical_models/smhm_models/zu_mandelbaum15.py#L108

PR #1031, opened 2022-07-01
Change obsolete call to astropy.core.cosmology to astropy.cosmology

aphearin:
Resolves FutureDeprecation warning in astropy


evevkovacs:
Thanks!

PR #1017, opened 2021-08-03
Adding GitHub Action to produce Python Wheels for Linux and Mac

EiffL:
This short PR adds a GitHub action that will automatically trigger the build of python wheels for all versions of python above 3.6 for both Mac (intel) and linux architectures. It uses the excellent https://cibuildwheel.readthedocs.io/ tool that makes all of this very simple.

For maintainers, here are the steps to follow upon a new code release:
- Use the github release tool to tag a new version
- Go get a coffee for like 30 mins
- Go the Actions tab and find the latest `Build wheels` run, download the attached `artifact`
- Unzip the artifact
- Upload to pypi with: `twine upload  *.whl`

This is in addition to uploading the normal source distribution. Eventually the entire process can be automatized if you create a pypi key so that the action can directly push to pypi.

After that, a user can trivially install halotools by:
```
$ pip install halotools
```
and they wont have to recompile anything, this will automatically grab the appropriate python wheel.

If you want to test it out before merging, I've run the action on my fork:
- The output of the github action is here (notice the attached artifact): https://github.com/EiffL/halotools/actions/runs/1094382382
- I pushed to test.pypi.org for testing: https://test.pypi.org/project/halotools/
- To try installing halotools with binary wheel from the test pypi, do the following:
```
$ pip install -i https://test.pypi.org/simple/ halotools
```

Annnnd to demonstrate how cool this is: https://colab.research.google.com/drive/1_KjhW8zV7EGDyIzeIHC9tZiZpffs69G7?usp=sharing

PR #1035, opened 2022-09-14
Update build wheel to macos-11

aphearin:
According to [this issue](https://github.com/actions/virtual-environments/issues/5583), wheels based on macos-10.15 are now deprecated. This updates the Action brought in by #1017 so that macos-11 is now used instead.

Tagging @EiffL in case he has any other suggestions for wheel updates. If not, this is the last planned commit prior to releasing v0.8.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-20.04, macos-10.15]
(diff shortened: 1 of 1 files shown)


EiffL:
Yeah I think that's fine. But also note that while you're at it, you should update the version of `cibuildwheel` to the latest release.
So, on line 21:
```
        uses: pypa/cibuildwheel@2.10.0
```

This will ensure that you build wheels for new versions of python as well, like 3.10 and even 3.11

PR #1041, opened 2022-09-15
Implement pyproject.toml

aphearin:
This WIP PR attempts to resolve RTD build failures.


aphearin:
@pllim Could you have a look at docs/conf.py and .readthedocs.yml to see if you understand why I would be getting this error when building the halotools docs on RTD
```ModuleNotFoundError: No module named 'numpy'```

**Comment by pllim** [comment]
@aphearin , could you please enable RTD PR build so we can debug this here?

https://docs.readthedocs.io/en/stable/pull-requests.html

On the GitHub side, under Settings, go to RTD webhook and make sure it is sending PR event. Then you can add/amend commit here and see if it pops up under the CI list. Thanks!

**Comment by aphearin** [comment]
Hmm, I enabled the webhook (checking Push to Repository and also Pull Request), and then I pushed up a forgotten change to tox.ini, but there doesn't seem to be a CI check that is showing up. However, on RTD, when I inspect the webhook, I see that this push shows up under Recent Activity.

**Comment by aphearin** [comment]
Well, hang on, forgot to hit save on the "build pull requests" option on the RTD admin page. Sorry, one sec...

**Comment by aphearin** [comment]
Ok, that did the trick - RTD build is now showing up as a CI check.

**Comment by pllim** [comment]
Looks like you are not using pyproject.toml yet. I wonder if your current setup is too old for RTD to understand. It did not seem to pick up Cython requirement from your requirements.txt . I would say look at how astroquery is doing it (https://github.com/astropy/astroquery/blob/main/.readthedocs.yaml) since it is also still using astropy-helpers and has a working RTD build.

**Comment by aphearin** [comment]
Ok, I can try and mirror what astroquery is doing. But is the migration to pyproject.toml that laborious? Do you think it would be better to just go ahead and handle that migration before releasing? Or will that require overhauling the package-template dependencies in a way that will be very time-consuming?

**Comment by aphearin** [comment]
Actually, astroquery is using pyproject.toml - thanks for the pointer I'll see if I can just mirror what they're doing and see how it goes

**Comment by pllim** [comment]
> But is the migration to pyproject.toml that laborious?

I would say a couple of hours but it depends on your familiarity with it.

> Do you think it would be better to just go ahead and handle that migration before releasing?

Probably. If RTD cannot build it, chances are you will have problem building wheel for it as well during release. Or your users will have problem installing it if their `pip` is too new or something (just wildly guessing here).

> Or will that require overhauling the helpers dependencies in a way that will be very time-consuming?

Probably not. Would be nice but it would require maybe a couple of days on-and-off working on it.

**Comment by aphearin** [comment]
@pllim I've taken a stab at migrating to pyproject.toml and I've tried to follow the astroquery example. Both my CI Tests and RTD workflows have the same error:

`ImportError: cannot import name 'setup' from 'astropy_helpers.setup_helpers' (/home/docs/checkouts/readthedocs.org/user_builds/halotools/checkouts/1041/astropy_helpers/astropy_helpers/setup_helpers.py)
`

My `setup.py` file is the same as [the one in astroquery](https://github.com/astropy/astroquery/blob/436245056c3e624b08abb0095a1bbec642ed1526/setup.py). I think [this is the line](https://github.com/astropy/astroquery/blob/436245056c3e624b08abb0095a1bbec642ed1526/setup.py#L16) that causes the issue with all 3 workflows. I thought this might be due to an out-of-date version of the helpers, although I can see in the RTD traceback that there is a submodule update invocation before this error appears. Any ideas for how to proceed?

**Comment by pllim** [comment]
Looks like astroquery has helpers at https://github.com/astropy/astropy-helpers/tree/d2a6304a3e801bc2cb053ccf1cf09a9f1e62036c while yours is at https://github.com/astropy/astropy-helpers/tree/fd80be7cf8698e2350a919ef7fc89871db0ba580 . You might need to update your submodule?

**Comment by aphearin** [comment]
Yes, thanks, updating to the same version of the helpers used by astroquery resolved the CI Tests (man, my helpers had gotten really out of date). So good news is halotools has now successfully migrated to `pyproject.toml`! That was really worth doing - this migration will be easy for my other packages now that I've done it for a pretty nontrivial example.

Bad news is that switching to `pyproject.toml` did not resolve the RTD fail. I still get the same error message:
`ModuleNotFoundError: No module named 'numpy'`

File: .readthedocs.yml [hunk 1]
```diff
@@ -14,4 +14,5 @@ python:
   install:
     - method: pip
       path: .
-    - requirements: docs/requirements.txt
+      extra_requirements:
+        - docs
```
aphearin, line 18: @pllim If I understand what this line in the rtd.yml file is doing, it should be instructing RTD to look in setup.cfg for the `-docs` section of `[options.extras_require]` in order to find the documentation build requirements. This seems to be the way it works [on this line](https://github.com/astropy/astroquery/blob/436245056c3e624b08abb0095a1bbec642ed1526/.readthedocs.yaml#L10) of the .readthedocs.yml file in astroquery (this section of the astroquery setup.cfg is [here](https://github.com/astropy/astroquery/blob/436245056c3e624b08abb0095a1bbec642ed1526/setup.cfg#L172)). But numpy appears in this section of my setup.cfg, so I'm not sure why my RTD build has this persistent error `ModuleNotFoundError: No module named 'numpy'`

setup.cfg
```diff
@@ -1,35 +1,73 @@
...
+tests_require =
+   pytest-astropy
+build_requires =
+   Cython
+
+
+[options.extras_require]
+test=
+  pytest-astropy
+docs=
```
aphearin, line 67: Here is the section where I _think_ I am declaring numpy as a docs build dependency. But apparently this is not working like I think it should.

.readthedocs.yml
```diff
@@ -14,4 +14,5 @@ python:
   install:
     - method: pip
       path: .
-    - requirements: docs/requirements.txt
```
pllim, line 17: Make sure you delete "docs/requirements.txt" and make sure it is not set that way in the Admin panel, just in case...

(Though not sure if that will fix your error or not)

setup.cfg
```diff
@@ -1,35 +1,73 @@
...
-author_email = ahearin@anl.gov
-license = BSD
-url = http://astropy.org/
-edit_on_github = True
-github_project = astropy/halotools
+
+[options]
+python_requires = >=3.9
+
+install_requires=
```
pllim, line 52: No idea if the whitespace is required but might as well be consistent.
```suggestion
install_requires =
```

setup.cfg
```diff
@@ -1,35 +1,73 @@
...
+   numpy
+   scipy
+   astropy>=5.0
+   requests
+   beautifulsoup4
+tests_require =
+   pytest-astropy
+build_requires =
+   Cython
```
pllim, line 61: I would move this to pyproject.toml

setup.cfg
```diff
@@ -1,35 +1,73 @@
...
+[options.extras_require]
+test=
+  pytest-astropy
+docs=
+  matplotlib
+  sphinx-astropy>=1.5
+  scipy
+  numpy
+  Cython
+  astropy>=5
```
pllim, line 73: These are either declared as build or install dependencies, so they are actually not needed here.
```suggestion
```

.readthedocs.yml
aphearin: Ooooh, yes! The admin panel was not changed when I switched to `pyproject.toml`. Now I'm pointing to `docs/conf.py` and leaving blank the requirements file. Hopefully that will do the trick.

pllim:
So... I just realized that astroquery does not have to worry about C code like you do. Perhaps it is not quite possible to modernize your build infrastructure without applying APE 17. We are not longer pushing updates to astropy-helpers and looks like the build error still bubbles up from there even after you update the helpers to version used by astroquery. 😿

pllim:
Oh, wait wait... maybe try this and see:

https://github.com/astropy/astropy/blob/3912916dad56920514ba648be400a5f82add041a/pyproject.toml#L2-L7

setup.cfg
aphearin: Or maybe even delete it now in light of your recent discovery of how astropy handles their `[build-system]` section...

aphearin: Your suggestion to have the `[build-system]` section include `"oldest-supported-numpy"` seems to have worked! Thanks so much for tracking this down!

aphearin:
I think the build requirements issue is now resolved thanks to the suggestion of @pllim to look at [these lines of the astropy pyproject.toml file](https://github.com/astropy/astropy/blob/3912916dad56920514ba648be400a5f82add041a/pyproject.toml#L2-L7) (and also the suggestion to fix the error in my Admin page of RTD, d'oh!).

I've made a few more successive failures, each time resolving the failure by making straightforward changes to my `conf.py`  by referring to the relevant line(s) of [the conf.py module in astroquery.](https://github.com/astropy/astroquery/blob/main/docs/conf.py)

pllim:
Looks like you need to not use astropy-helpers in `docs/conf.py`. Instead use https://github.com/astropy/sphinx-astropy

docs/conf.py
```diff
@@ -153,7 +153,7 @@
 ## -- Options for the edit_on_github extension ----------------------------------------
 
 if eval(setup_cfg.get("edit_on_github")):
-    extensions += ["astropy.sphinx.ext.edit_on_github"]
+    extensions += ["sphinx_astropy.ext.edit_on_github"]
```
aphearin, line 156: @pllim you just suggested:
> Looks like you need to not use astropy-helpers in docs/conf.py. Instead use https://github.com/astropy/sphinx-astropy
After this change shown here, I no longer get the error that I have been getting (`No module named 'astropy_helpers.sphinx'). I think this is what you were suggesting all along, so if I understood you correctly, then I think you were right.

pllim: There are also imports in `docs/conf.py`.

aphearin: Right. Fixing that now.

aphearin: Actually, now I'm seeing that all of this work is going to make it easier later to migrate away from the helpers altogether. Previously that seemed like a really big lift, but actually it's really helpful that much of that functionality has been migrated to standalone packages such as sphinx_astropy.

aphearin:
Now that I am using `sphinx_astropy` within `docs/conf.py`, I have now reached the final line of `conf.py`, so with fingers crossed, this is the final bug to resolve:
```
Exception occurred:
  File "/home/docs/checkouts/readthedocs.org/user_builds/halotools/envs/1041/lib/python3.9/site-packages/halotools/utils/autosummary_workaround.py", line 6, in <module>
    class FixedAutosummary(Autosummary):
NameError: name 'Autosummary' is not defined

```

This bug can be traced to [this line in the halotools source code](https://github.com/astropy/halotools/blob/3449072811aa1044ef65490f7ae8e975c89dd312/halotools/utils/autosummary_workaround.py#L6), which was written by Erik way back when for reasons I forget. The `FixedAutosummary` class inherits from `Autosummary`, which appears to have been previously located somewhere in the `sphinx.ext.autosummary` namespace, but I guess not in more recent versions of sphinx?

I have grepped for `Autosummary` within the source code of [the sphinx-astropy package](https://github.com/astropy/sphinx-astropy), but I do not see where this is defined.

docs/conf.py
aphearin: I have now eliminated all calls to `astropy_helpers` within `conf.py`, and the docs build now appears to successfully proceed all the way to the final line of `conf.py`

pllim:
I would just ditch `FixedAutosummary` altogether and see if the doc rendering is still acceptable. All the Astropy projects use the vanilla autosummary and no one complained (that much).

pllim:
Oh yeah, I think you definitely don't need it. Sphinx is like v6 now.

https://github.com/astropy/halotools/blob/3449072811aa1044ef65490f7ae8e975c89dd312/halotools/utils/autosummary_workaround.py#L111

aphearin:
Whoa, yeah, good catch. Fingers crossed this does the trick.

aphearin:
Well I think that ditching the autosummary_workaround have been all that needs to be done on the packaging side. Unfortunately, I now have a ton of new errors showing up, presumably because some choices I made a while back pertained to grossly out-of-date version of sphinx, and these show up now that I've updated all the dependencies. So I think I still have a lot of work to do, but at least now the problems don't seem to have anything to do with packaging, only my docstring formats. I'll try and tackle the reformatting next week - thanks so much for all the help @pllim!

aphearin:
It turns out that the docstring formatting issues weren't bad at all - only a single failure from a missing colon, and the rest were just warnings with simple fixes. Reiterating many thanks to @pllim for all the patience and support!

PR #1048, opened 2022-09-25
Resolve cibuildwheel failures

aphearin:
In PR #1047 halotools removed dependency on the astropy-helpers and began using extensions-helpers to handle the automation of the cython extensions. In the current master branch, [the cibuildwheel github action](https://github.com/astropy/halotools/blob/master/.github/workflows/wheels.yml) in current master branch [fails to successfully build a wheel](https://github.com/astropy/halotools/actions/runs/3122524070/jobs/5064519017) to upload to pypi. This PR attempts to resolve this failure.


aphearin:
The build wheel failures were due to an older version of cython. This was resolved by updating cython from `0.29.14` to `0.29.32`

PR #1063, opened 2023-06-11
Implement fix for inconsistent treatment of littleh in Leauthaud+11 model

### LINKED ISSUE #1061: Leauthaud11 HOD
Within the Leauthaud11 HODs internal calculations, can I make halotools use h=1 units all the way through? For instance, the Leauthaud11 parameters like `smhm_m0_0`, `smhm_m1_0`, and `bsat` are set to default values from the literature in the units of h=0.7 or 0.72. So, while the user inputs like the stellar mass threshold to initialize the Leautahaud11 HOD model are given in the h=1 units, halotools internally converts the stellar mass threshold to the h=0.7 or 0.72 units, matching the literature values of the other Leauthaud11 HOD parameters. Then it does the calculations and returns the output/s in h=1 units e.g within the ```mean_log_halo_mass()``` and ```_update_satellite_params()``` functions within ``halotools.empirical_models.occupation_models.leauthaud11_components```. 

Moreover, within the aforementioned `_update_satellite_params()` function, in the following line, `knee_threshold = (10.0**log_halo_mass_threshold) * self.littleh`, should the `self.littleh` be divided and not multiplied to bring the `knee_threshold` to the same footing (h=0.72) as `kness_mass` coming from the literature?

aphearin:
This PR resolves #1061 as follows:

- The new Leauthaud11 has parameter values that assume h=0.72, so that the numerical values in the parameter dictionary can be directly compared to Leauthaud+11.
- The input stellar mass threshold is quoted in h=1, so that the user needs to rescale the stellar mass of their sample before directly comparing parameter values.
- All user-facing functions such as `mean_stellar_mass` and `mean_log_halo_mass` accept and return values of mass in h=1 units, for consistency with the rest of the package.
- There is no longer any dependence upon Behroozi10SmHm. Instead, there is a new Leauthaud11SmHm class defined within the same module. This simplifies the complication that Behroozi+10 assumed h=0.7, whereas Leauthaud+11 assumed h=0.72.

Throughout the docstrings, I have made an effort to write copious, very explicit notes about how to convert back and forth between halo and stellar mass.

Thanks to @zaidikumail for the patience. Comments from @zaidikumail in particular would be most welcome.

### DIFF
--- a/halotools/empirical_models/occupation_models/leauthaud11_components.py
+++ b/halotools/empirical_models/occupation_models/leauthaud11_components.py
@@ -10,11 +10,18 @@ import warnings
 from .occupation_model_template import OccupationComponent
 
 from .. import model_defaults
-from ..smhm_models import Behroozi10SmHm
 from ..assembias_models import HeavisideAssembias
 
 from ... import sim_manager
 from ...custom_exceptions import HalotoolsError
+
+from ..component_model_templates import PrimGalpropModel
+from ..smhm_models.smhm_helpers import safely_retrieve_redshift
+from .. import model_helpers
+
+
+L11_LITTLEH = 0.72
+L11_LGH = np.log10(L11_LITTLEH)
 
 __all__ = (
     "Leauthaud11Cens",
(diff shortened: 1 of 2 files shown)


zaidikumail (approved):
I reviewed the changes you made @aphearin. Everything looks great now. The little h usage is consistent throughout and thanks for explicitly labeling the different quantities with `...h0p72` or `..h1p0` to denote their little h footings. Thanks for also fixing the errors in the satellite occupation part, for instance, in lines 392-393 where the correctly calculated `halo_mass_h0p72` is being used.

aphearin:
Great, thanks for taking the time to carefully review the changes @zaidikumail (and for raising this issue in the first place). Much appreciated!

FYI - I have one more bug I'd like to fix and then I plan to do a bugfix release 0.8.2, hopefully as soon as next week.

PR #1072, opened 2023-10-08
Update readthedocs.yml

aphearin:
Update readthedocs.yml


pllim:
Already fixed, I think.

PR #1080, opened 2024-01-11
Typo in tpcf_jackknife.py?

ShuleiCao:
In line 416--418 of halotools/mock_observables/two_point_clustering/tpcf_jackknife.py, np.min(x3), np.min(y3), and np.min(y3) are supposed to be np.max(x3), np.max(y3), np.max(y3), right?


aphearin:
Ooh, yes, good catch @ShuleiCao that on this bug. Thanks for the fix!

PR #1082, opened 2024-07-31
Fix crashes due to rounding errors in mean_occupation functions

### LINKED ISSUE #1081: Crash in mock generation due to rounding errors
The following code creates an error. Note that all numbers matter. Different choices for model parameters likely won't produce the error.

```
from halotools.empirical_models import PrebuiltHodModelFactory
from halotools.sim_manager import CachedHaloCatalog

model = PrebuiltHodModelFactory(
    'cacciato09', threshold=8.5, prim_haloprop_key='halo_mvir',
    prim_galprop_key='luminosity')
model.param_dict['log_M_1'] = 11.2
model.param_dict['log_L_0'] = 9.95
model.param_dict['gamma_1'] = 3.5
model.param_dict['gamma_2'] = 0.25
model.param_dict['a_1'] = 0.8
model.param_dict['a_2'] = 0.0

halocat = CachedHaloCatalog(simname='bolplanck', redshift=0,
                            halo_finder='rockstar')
model.populate_mock(halocat)
```

```

  File ~/.venv/lib64/python3.12/site-packages/spyder_kernels/py3compat.py:356 in compat_exec
    exec(code, globals, locals)

  File ~/Projects/halotools/scratch.py:16
    model.populate_mock(halocat)

  File ~/.venv/lib64/python3.12/site-packages/halotools/empirical_models/factories/hod_model_factory.py:1209 in populate_mock
    ModelFactory.populate_mock(self, halocat, **kwargs)

  File ~/.venv/lib64/python3.12/site-packages/halotools/empirical_models/factories/model_factory_template.py:254 in populate_mock
    self.mock.populate(**mockpop_kwargs)

  File ~/.venv/lib64/python3.12/site-packages/halotools/empirical_models/factories/hod_mock_factory.py:326 in populate
    self.allocate_memory(seed=seed)

  File ~/.venv/lib64/python3.12/site-packages/halotools/empirical_models/factories/hod_mock_factory.py:495 in allocate_memory
    self.galaxy_table[halocatkey] = np.zeros(

ValueError: negative dimensions are not allowed
```

The underlying reason seems to be that the Cacciatio09 satellite occupation model produces an extremely small negative mean expected occupation for some halos. This is likely due to some rounding error since the value is something like -3e-47. This negative expectation value than correctly creates an NaN when calling `pdtrik` [here](https://github.com/astropy/halotools/blob/769ed1dd470622d177a161cd1866539ac741177e/halotools/empirical_models/occupation_models/occupation_model_template.py#L210). However, in the same part of the code, this NaN is then explicitly converted into an in int which is -9223372036854775808. Ultimately, the expected galaxy table length is then negative, creating the crash.

Here's a potential solution: Right before calling `pdtrik` to create draws from a Poisson distribution, we should check whether expectation values are negative. If so but they're effectively 0 to within some precision, we could force negative expectation values to be 0 exactly. If however, some values are negative beyond some reasonable value for numerical noise, the code should throw an error to alert the user that the model produced an unphysical occupation. @aphearin I can write this if you think that's a good solution.

johannesulf:
This PR adds a check that the mean occupation for all halos is non-negative. If it's negative but reasonably close to 0 such that this may be caused by rounding errors, the mean occupation will be 0 exactly. If it's clearly negative, `halotools` will raise an error to alert the user that the model is unphysical. This should fix #1081.

@aphearin This PR also adds the above check for centrals and satellites. Let me know if this PR could produce any unwanted exceptions.

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -40,11 +40,11 @@ jobs:
 
     steps:
     - name: Checkout code
-      uses: actions/checkout@v2
+      uses: actions/checkout@v4
       with:
         fetch-depth: 0
     - name: Set up python ${{ matrix.python }} on ${{ matrix.os }}
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.python }}
     - name: Install base dependencies
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -44,7 +44,7 @@ jobs:
       with:
         fetch-depth: 0
     - name: Set up python ${{ matrix.python }} on ${{ matrix.os }}
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@v4
       with:
         python-version: ${{ matrix.python }}
     - name: Install base dependencies
--- a/halotools/empirical_models/occupation_models/occupation_model_template.py
+++ b/halotools/empirical_models/occupation_models/occupation_model_template.py
@@ -135,6 +135,18 @@ class OccupationComponent(object):
             Integer array giving the number of galaxies in each of the input table.
         """
         first_occupation_moment = self.mean_occupation(**kwargs)
+
+        if np.any(first_occupation_moment < 0):
+            if np.all(first_occupation_moment >= -1e-6):
+                first_occupation_moment = np.maximum(
+                    first_occupation_moment, 0)
+            else:
+                msg = (
+                    "\nThe mean occupation for at least one halo was negative. \n"
+                    "Ensure that the mean occupation is always non-negative. \n"
+                )
+                raise HalotoolsError(msg)
+
         if self._upper_occupation_bound == 1:
             return self._nearest_integer_distribution(
                 first_occupation_moment, seed=seed, **kwargs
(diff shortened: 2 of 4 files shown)


aphearin:
Thanks to @johannesulf and Kaustav for a simple fix of this subtle issue. I'm not sure why CI is failing at the moment, but it looks like it doesn't have anything to do with the code in the PR.

aphearin:
Well this workflow failure appears to be persistent (sometimes fails like this are only transient due to a temporary issue in the dependency chain). I found [this github page](https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/) indicating that the problem has something to do with an outdated workflow configuration, and so I tried simply updating checkout@v2 to checkout@v4 but that did not resolve the issue.

johannesulf:
Sorry for taking a bit longer on this. I suspect that some of the workflow failures may be related to numpy 2. I get similar errors if I don't specify numpy==1.26.4. Ultimately, I was able to run the tests. Two small modifications to the tests had to be made.
- One tests assumed that the first occupation moment can be `None`, initially. My code for `mc_occupation` didn't allow that. But since that behavior seems like a quirk, I just slightly changed the test so that the first occupation moment isn't `None`.
- The recently released scipy 1.14 finally deprecated `scipy.integrate.cumtrapz`. So one tests needed to be updated to accommodate that change.

aphearin:
Thanks for the leg work on code dependency issues. I'll try and solve the CI fails separately in #1083 but this looks good to me to go ahead and merge in to master.

PR #1083, opened 2024-08-14
Add new cron workflow

aphearin:
Add new cron workflow that runs CI tests as a cron job every month

### DIFF
--- a/.github/workflows/cron_tests.yml
+++ b/.github/workflows/cron_tests.yml
@@ -1,13 +1,7 @@
 name: Cron Tests
 
 on:
-  push:
-    branches:
-    - master
-  #  tags: # run CI if specific tags are pushed
-  pull_request:
-    branches:
-      - master
+  workflow_dispatch: null
   schedule:
     # Runs "First of every month at 3:15am Central"
     - cron: '15 8 1 * *'
--- a/.github/workflows/cron_tests.yml
+++ b/.github/workflows/cron_tests.yml
@@ -1,0 +1,59 @@
+name: CI Tests
+
+on:
+  push:
+    branches:
+    - master
+  #  tags: # run CI if specific tags are pushed
+  pull_request:
+    branches:
+      - master
+  schedule:
+    # Runs "First of every month at 3:15am Central"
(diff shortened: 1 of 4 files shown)


aphearin:
This PR brings in a new cron workflow that will run in the background every month. Our CI currently fails for some reason due entirely to workflow config, and so #1082 can now be merged even though it is failing.

aphearin:
As reported in #1082, the halotools CI now fails in a way that is resolved by only running tests with `numpy==1.26.4`. The last two commits of this PR (https://github.com/astropy/halotools/pull/1083/commits/d82cdc5c0e85ffe037684bcf2dac676ec9ef918f and https://github.com/astropy/halotools/pull/1083/commits/a5b0b59f47a508ab4398fc174aa81987124b6802) attempt to modify our CI so that `numpy<2` is used, but these attempts did not succeed at creating the correct environment, because the following message still appears in the test report:

```
------------------------------- Captured stderr --------------------------------
A module that was compiled using NumPy 1.x cannot be run in
NumPy 2.1.0 as it may crash. To support both 1.x and 2.x
versions of NumPy, modules must be compiled with NumPy 2.0.
Some module may need to rebuild instead e.g. with 'pybind11>=2.12'.

If you are a user of the module, the easiest solution will be to
downgrade to 'numpy<2' or try to upgrade the affected module.
We expect that some modules will need time to support NumPy 2.
```

@johannesulf when you said you ran tests successfully with `numpy==1.26.4`, were you referring to your local installation? Or tests run through CI on github?

aphearin:
Hey nvm @johannesulf, I was just adding `numpy<2` to the wrong place to communicate with tox. It was super helpful that you identified the root source of the failure was future compatibility with numpy - thanks again!

aphearin:
Besides bringing in a new cron workflow, this PR resolves the workflow failures that prevented CI from running on #1082. The resolution was to change install requirements to `numpy<2`. It's not clear whether this is a compatibility issue between halotools source code and `numpy>=2`, or some incompatibility elsewhere in our dependency chain that will resolve in time.

PR #1095, opened 2024-09-09
Fix readthedocs formatting issues

aphearin:
As noticed by @pllim a recent PR (probably #1093) broke the readthedocs formatting. This PR attempts to fix.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -7,7 +7,7 @@ from datetime import datetime, timezone
 from importlib import metadata
 
 # Load all of the global Astropy configuration
-from sphinx_astropy.conf import exclude_patterns
+from sphinx_astropy.conf import *
 
 # -- General configuration ----------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


pllim:
Nope... See rendered doc at https://halotools--1095.org.readthedocs.build/en/1095/

pllim:
Maybe you are not using the theme that you think you are using. Raw log might give you more info. Example: https://readthedocs.org/api/v2/build/25571238.txt

pllim:
BTW, if you want to save CPU cycles for CI, you can put `[ci skip]` on your commit message. Actions respects that and won't run. RTD does not, so it would run anyway.

pllim:
Looks good now. Glad you figured it out! 🎉 

https://halotools.readthedocs.io/en/latest/

aphearin:
Yes! Mysteriously, it was just this one-liner: e94ce2aea7d87f21d830c98c93718451c40342e9. Thanks again for all the help @pllim!

PR #1111, opened 2024-12-18
Update requirements to python>=3.11

nvanalfen:
Updated the required Python version in the installation documentation to Python 3.10 or later as installation fails otherwise.

### DIFF
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -18,15 +18,6 @@ jobs:
     strategy:
       matrix:
         include:
-          - name: Ubuntu - Python 3.10 with all optional dependencies
-            os: ubuntu-latest
-            python: "3.10"
-            toxenv: 'py10-test-alldeps'
-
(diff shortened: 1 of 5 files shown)


aphearin:
Can you just increment to python>=3.11?

nvanalfen:
Sure. Just confirming, we want to list 3.11? It works with 3.10, but should I just change that recommended installation piece to >=3.11 and leave the installation requirement as"3.10 or later"?

aphearin:
Well, the stated installation requirement should always match the actual installation requirement, whatever it is.

I have been meaning to drop support for 3.10, and now is as good of a time as any. So could you update the requirement as well as the statement?

nvanalfen:
Got it. Will do. Thanks for confirming

aphearin:
@nvanalfen could you also modify `pyproject.toml` to mandate python>=3.11, and also the CI workflows?

aphearin:
Closed by #1114

PR #1114, opened 2025-01-27
Add support for numpy 2

AlanPearl:
This pull request brings halotools up to date with numpy 2 to bring #1084 to a close. Build system now requires numpy>=2, which compiles code compatible with both numpy 1 and 2.

Edit: Might still need to replace `oldest-supported-numpy` with `numpy>=2` [here](https://github.com/astropy/halotools/blob/master/.github/workflows/wheels.yaml#L30)?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ classifiers = [
     "Programming Language :: Python :: 3",
     "License :: OSI Approved :: BSD License",
 ]
-requires-python = ">=3.10"
+requires-python = ">=3.11"
 
 [tool.setuptools.dynamic]
 dependencies = {file = ["requirements.txt"]}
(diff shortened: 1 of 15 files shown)


pyproject.toml
```diff
@@ -51,8 +51,8 @@ markers = ["slow", "installation_test"]
 requires = ["setuptools>=42.0.0",
             "setuptools_scm",
             "wheel",
-            "oldest-supported-numpy",
```
aphearin, line 54: @bsipocz this PR led by @AlanPearl establishes compatibility between halotools and numpy 2.0. However, I'm not sure about this particular line of code because I never really understood what using `oldest-supported-numpy` did. But this PR fails unless this line is changed to `numpy>=2.0`.

On a possibly related note, I'm also not sure whether any additional changes need to be made in order to properly use `extension-helpers` in the right way.

PR #1123, opened 2025-02-18
Trivial change to readme to re-trigger test runners

aphearin:
#1122 may have broken something based on [this failed test report](https://github.com/astropy/halotools/actions/runs/13384479632), even though all the PR checks passed before merging. The purpose of this trivial PR is just to re-trigger the full CI test suite.


pllim:
You might wanna grep the code base for more mentions of `.test()` and get rid of them too. FYI.

https://github.com/search?q=repo%3Aastropy%2Fhalotools%20.test&type=code

pllim:
And I don't think you can do this anymore: https://github.com/astropy/halotools/blob/master/halotools/__init__.py

aphearin:
Whoops, great, thanks @pllim for catching this! It looks like this was not caught by my CI.

aphearin:
@pllim Do you know whether there is any need to keep the [conftest.py](https://github.com/astropy/halotools/blob/master/halotools/conftest.py) module anymore? If the only purpose was to enable astropy test runners, then maybe this module can be removed wholesale?

pllim:
conftest.py is used by pytest. Without the test runner, you could move it up to the root directory if you want. The stuff you currently have configures the pytest header. But you can add other stuff to it in the future as needed as well.

Example: https://docs.pytest.org/en/stable/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files

PR #1124, opened 2025-02-18
Count cic indexes with int64 to prevent overflow

AlanPearl:
This PR will provide a solution to the ValueError in galtab [issue 6](https://github.com/AlanPearl/galtab/issues/6) and does not degrade performance. I'll leave the example test script here, since it is too expensive to include as a unit test:

```
# NOTE Memory requirement: ~46 GB
import numpy as np
import halotools.mock_observables as htmo

sample = np.zeros((40_000, 3))
htmo.counts_in_cylinders(sample, sample, 1, 1, return_indexes=True, period=100)
```


aphearin:
Thanks for the fix @AlanPearl!

PR #1130, opened 2025-08-04
Extend redshift of `ra_dec_z` function from z=1 to z=20

aphearin:
Extend redshift of `ra_dec_z` function from z=1 to z=20


aphearin:
Thanks to @sortegamtnez for catching this!

